PR #44, opened 2014-11-14
Introducing Remote operation to fetch users' quota

MMMarcy:
As the title said this simpli add a remote user operation.
Depends on https://github.com/owncloud/android-library/pull/38


src/com/owncloud/android/lib/resources/users/GetRemoteUserQuotaOperation.java
```diff
@@ -0,0 +1,119 @@
...
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+
+/**
+ * Created by marcello on 11/11/14.
+ */
+public class GetRemoteUserQuotaOperation extends RemoteOperation {
+
+    private static final String TAG = GetRemoteUserNameOperation.class.getSimpleName();
```
davivel: Typo, getting the name from other class

src/com/owncloud/android/lib/resources/users/GetRemoteUserQuotaOperation.java
```diff
@@ -0,0 +1,119 @@
...
+    @Override
+    protected RemoteOperationResult run(OwnCloudClient client) {
+        RemoteOperationResult result = null;
+        int status = -1;
+        GetMethod get = null;
+
+
+        //Get the user
+        try {
+            OwnCloudBasicCredentials credentials = (OwnCloudBasicCredentials) client.getCredentials();
```
davivel: This makes the solution available only for servers with authentication though Basic Auth; need to check the code of OwnCloudCredentials and classes extending it, I won't dare to purpose an alternative from memory.

src/com/owncloud/android/lib/resources/users/GetRemoteUserQuotaOperation.java
```diff
@@ -0,0 +1,119 @@
...
+            String url = client.getBaseUri() + OCS_ROUTE + credentials.getUsername();
+
+            get = new GetMethod(url);
+            get.addRequestHeader(createAuthenticationHeader(credentials));
```
davivel: This line not necessary, OwnCloudClient applies the credentials. In this point of the code it should be already authenticated.

src/com/owncloud/android/lib/resources/users/GetRemoteUserQuotaOperation.java
```diff
@@ -0,0 +1,119 @@
...
+        }
+
+        return result;
+    }
+
+    private boolean isSuccess(int status) {
+        return (status == HttpStatus.SC_OK);
+    }
+
+    private Header createAuthenticationHeader(OwnCloudCredentials credentials){
```
davivel: This method is not needed.

davivel:
Seen, need some changes. I'll have a look to the issue of getting the username no matter the credentials type.

In any case, I wouldn't make any change until android#737 is reviewed also.

davivel:
owncloud/android#737 was finally reviewed; please take a look at the comments in both

malkomich:
Let's track this PR here: https://github.com/owncloud/android-library/pull/102

PR #50, opened 2015-01-23
Fix encoding

RealRancor:
This fixes problems when building the lib from source like:

```
    [javac] /usr/src/oc-android-library/src/com/owncloud/android/lib/resources/shares/CreateRemoteShareOperation.java:69: warning: unmappable character for encoding UTF-8
    [javac] 	 * @param shareType			�0� = user, �1� = group, �3� = Public link. Mandatory argument
```


RealRancor:
Sorry, sometimes it could be useful to read the README first. This is also already fixed in the develop branch with the commit: https://github.com/owncloud/android-library/commit/04299bdf3c03b860254247377f2d03d5eb446280

PR #55, opened 2015-03-05
Update propfind for server 8.1 (WIP)

masensio:


src/com/owncloud/android/lib/resources/files/ReadRemoteFolderOperation.java
```diff
@@ -72,9 +75,27 @@ protected RemoteOperationResult run(OwnCloudClient client) {
         PropFindMethod query = null;
         
         try {
-            // remote request 
+            // PropFind Properties ( instead of DavConstants.PROPFIND_ALL_PROP )
+            DavPropertyNameSet propSet = new DavPropertyNameSet();
+            propSet.add(DavPropertyName.DISPLAYNAME);
```
davivel: We should create the propSet in a common method to be called by both operations.

davivel:
:+1:

PR #56, opened 2015-03-20
Add user agent support

masensio:


DeepDiver1975:
Can you also please add the version? THX

DeepDiver1975:
Furthermore (quoting @danimo) a generic browser user agent is required to not be blocked by firewalls.

danimo:
The user agent in the desktop client is as follows:

````Mozilla/5.0 (%1) mirall/%2```

where `%1` is the platform (Windows, Linux, MacOS), and `%2` is the version.

Example: ````Mozilla/5.0 (MacOS) mirall/1.8.0```

Rationale: Some Layer 7 firewalls use the "Mozilla.." part as plausibility check.

We could probably add an `(ownCloud)` as a common identifier, but I am not sure about wether OEMs would like that.

**Comment by DeepDiver1975** [comment]
we for sure shall apply identical user agent strings on the ios app as well.

@LukasReschke we use some user agent in core as well when accessing external resources - we might get affected by the firewall stuff as explained above by @danimo ???

**Comment by davivel** [comment]
@masensio , let's replace the value witg

Mozilla/5.0 (Android) ownCloud /1.7.0

@danimo, apart from the Layer 7 firewalls check, any other reason that justifies the user of "Mozilla"? For instance, does the desktop client use any network library related to Mozilla Foundation?

File: src/com/owncloud/android/lib/common/operations/RemoteOperation.java [hunk 1]
```diff
@@ -90,23 +96,29 @@
...
      * 
-     * @param account   ownCloud account in remote ownCloud server to reach during the execution of the operation.
+     * @param account   ownCloud account in remote ownCloud server to reach during the
+     *                  execution of the operation.
      * @param context   Android context for the component calling the method.
+     * @param userAgent userAgent string
      * @return          Result of the operation.
      */
-    public RemoteOperationResult execute(Account account, Context context) {
+    public RemoteOperationResult execute(Account account, Context context, String userAgent) {
```
davivel: We can't do this, need to find other solution. If the user agent is set in the OC client, we shouldn't need to include it every execute() call; that redundancy in parameters should not be needed.

src/com/owncloud/android/lib/common/OwnCloudClient.java
```diff
@@ -207,7 +209,11 @@ public int executeMethod(HttpMethod method) throws IOException, HttpException {
         		*/
 	            customRedirectionNeeded = mFollowRedirects;
 	        }
-        
+
+            // Update User Agent
+            HttpParams params = method.getParams();
+            params.setParameter(HttpMethodParams.USER_AGENT, mUserAgent);
```
davivel: This should be enough. For the buggy GET operations detected we may apply the specific code that's needed, but all the others should need nothing else.

masensio:
The userAgent string is:
Mozilla/5.0 Android-ownCloud /n
n: version number

rperezb:
@danimo where are you including the app name? we need this, don't we @MTRichards ?

davivel:
:+1:

danimo:
Why not add the app version?

masensio:
Sorry, maybe the description 
Mozilla/5.0 (Android) ownCloud /n
n: version number
is confused. /n isn't an intro.

For the actual version user-agent is:
Mozilla/5.0 (Android) ownCloud/1.7.0
where the version is 1.7.0

MTRichards:
@danimo @karlitschek  
Question.
Should the agent string use Mozilla, which eases firewall pass through, or should it use the app network stack. For example:
Apache HTTP Client/x (Android) ownCloud/1.7.0
iOS Networking Library/x (iOS) ownCloud/3.1.2

OR

Mozilla/5.0 (Android) ownCloud/1.7.0
Mozilla/5.0 (iOS) ownCloud/3.1.2

OR a mix of the two

don't know enough about the proper protocol here, looking for guidance.

davivel:
@MTRichards , after some extra research, Mozilla/5.0 inclusion should not be a problem.

@danimo explained the reason to ownCloud use it in https://github.com/owncloud/client/issues/646 .

MTRichards:
Ok, then we do this:
Mozilla/5.0 (Android) ownCloud/1.7.0
Mozilla/5.0 (iOS) ownCloud/3.1.2

PR #57, opened 2015-03-23
Create constants for share permission

przybylski:


przybylski:
@davivel any thoughts ?

davivel:
Sorry, but currently we don't accept PRs with purely cosmetic purposes. The load of prioritizing and reviewing PRs could increase dramatically if we opened that door.

The code seems fine, but it's not adding any functionality. I'd like to redirect you to some issue in the app repository regarding to handling permissions, but I cannot find any right now. We can discuss the functionality to use these changes in this repository or in the app repo.

Feel free to extend the work in this sense (not too much, huge PRs are still worse that tiny ones) and we'll reconsider.

przybylski:
This was suppose to be part of bigger changeset which integrates android wear into android app, but its not part of android-library.

Also explenation of x+y+z in comment is misleading if someone doesn't understand how oC is informing about share type.

PR #63, opened 2015-05-11
Fix refresh large default storage

masensio:
Adapt WebDav parser for receiving exponential values in QUOTAS


davivel:
:+1: , waiting for QA validation

davivel:
@purigarcia approved also, going in.

PR #65, opened 2015-05-26
Fix problems reading quota values: NullPointerException

masensio:


davivel:
:+1:

davivel:
Going into after QA's OK

PR #67, opened 2015-06-02
Use .equals instead of !=

LukasReschke:
This code never worked as intended. Please consider adding unit tests.

### DIFF
--- a/src/com/owncloud/android/lib/common/accounts/AccountUtils.java
+++ b/src/com/owncloud/android/lib/common/accounts/AccountUtils.java
@@ -218,7 +218,7 @@ public class AccountUtils {
 
 		if (client != null) {
 			String cookiesString = client.getCookiesString();
-			if (cookiesString != "") {
+			if (!cookiesString.equals("")) {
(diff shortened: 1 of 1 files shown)


src/com/owncloud/android/lib/common/accounts/AccountUtils.java
```diff
@@ -218,7 +218,7 @@ public static void saveClient(OwnCloudClient client, Account savedAccount, Conte
 
 		if (client != null) {
 			String cookiesString = client.getCookiesString();
-			if (cookiesString != "") {
+			if (!cookiesString.equals("")) {
```
PVince81: Make this `!"".equals(cookiesString)` because if cookiesString happens to be null, you'll get `NullPointerException`

LukasReschke: Done - thanks :smile:

davivel:
Thanks, pals. :+1:

For the futture, please, PRs on this repository *must* go into develop, not into master branch.

PR #68, opened 2015-06-02
Close resource in finally block

LukasReschke:
I don't quite get the reason for `mBuf` here as in `com.owncloud.android.lib.common.utils.Log_OC#appendLog` anyways a new `BufferedWriter` is started but it might make sense to close this here as this will otherwise cause warnings in static code scanners.

Didn't test this as I don't have a Android device and so it requires extensive testing.


davivel:
:+1:

davivel:
Thanks for cleaning, @LukasReschke; my fault.

PR #74, opened 2015-06-23
Added check of support of remote thumbnails to OwnCloudVersion

davivel:


DeepDiver1975:
For the sake of a clean git history I advice to squash the commits.

9 changed lines in 2 files but 4 commits looks strange

davivel:
How do you use to do in core. Squish all the commits in every PR?

DeepDiver1975:
> How do you use to do in core. Squash all the commits in every PR?

Generally speaking one commit is a selfcontained unit which in itself is compiling and not introducing issues (as in co is not failing).

For bigger changes exceptions are valid - it's pointless to have a few k of source changes in one commit - but reasonable commits are possible as well.

Finally: if you use rebase instead of merge you can avoid such merge commits which are ugly as well.

davivel:
Yeah, we considered rebase several times in the past; problem was having to deal with conflict propagated through successive stages of the rebase process.

Will give a it a try anyway. Thanks.

davivel:
Thanks for the suggestion, @DeepDiver1975 . It even helped with the unit tests. You rock.

DeepDiver1975:
> Thanks for the suggestion, @DeepDiver1975 . It even helped with the unit tests. You rock.

My pleasure - feel free to ping me any time in case you have question.

PR #77, opened 2015-07-04
Allow asynchronous execution without handler

mendhak:
Currently `RemoteOperation.execute` requires a `Handler` so that it can post about an operation being finished - this almost always requires it being invoked from an activity or service. 

When invoking this from an isolated background thread, there is no `Handler` available.  This `execute` overload removes the `Handler` argument so that it can be called from such background threads. 

Reference: Issue #64

PR #95, opened 2015-11-10
Mendhak master

DeepDiver1975:
non fork pr of https://github.com/owncloud/android-library/pull/77 to get travis green

### DIFF
--- a/src/com/owncloud/android/lib/common/operations/RemoteOperation.java
+++ b/src/com/owncloud/android/lib/common/operations/RemoteOperation.java
@@ -254,6 +254,34 @@ public abstract class RemoteOperation implements Runnable {
 		runnerThread.start();
 		return runnerThread;
 	}
+
+    /**
+     * Asynchronously executes the remote operation without a handler
+     *
+     * @param client			Client object to reach an ownCloud server
+     *                          during the execution of the operation.
+     * @param listener			Listener to be notified about the execution of the operation.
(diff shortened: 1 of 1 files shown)


src/com/owncloud/android/lib/common/operations/RemoteOperation.java
```diff
@@ -254,6 +254,34 @@ public Thread execute(OwnCloudClient client,
...
+
+    /**
+     * Asynchronously executes the remote operation without a handler
+     *
+     * @param client			Client object to reach an ownCloud server
+     *                          during the execution of the operation.
+     * @param listener			Listener to be notified about the execution of the operation.
+     * @return					Thread were the remote operation is executed.
+     */
+    public Thread execute(OwnCloudClient client,
```
przybylski: Well... this is almost copy paste from the other execute. Couldn't we just soften the requirements on the other execute to allow handler to be null? It is checked anyway before usage for null.

przybylski: Or even better, why should RemoteOperation care what it is calling back to? It should just callback to listener, its the listener responsibility to correct handle callback. Bu that depends on numbers of usages ;)

davivel:
@DeepDiver1975 , @mendhak resumed the work in his original PR #77. Will go on there.

Thanks!

PR #98, opened 2015-11-23
Changes to support update of password and expiration time in public shares

davivel:


davivel:
Ready to review.

Last commit from @masensio is :+1: from my side; @masesnsio, confirm the rest.

Any other reviewers are welcome.

rperezb:
:+1:

PR #100, opened 2015-12-09
Fix bug related with sharee with special characters

masensio:


davivel:
:+1:

(Aligning master with stable after release)

PR #101, opened 2015-12-20
Correctly setup lint options

przybylski:


davivel:
Right. If you don't mind, I will rebase by myself.

:+1:

davivel:
@jesmrec , no QA needed here.

PR #108, opened 2016-02-10
Update master with release 0.9.10

davivel:


davivel:
Self-merge as special PR to update released code.

PR #112, opened 2016-03-14
Support for changes in https://github.com/owncloud/android/pull/1493

davivel:
@owncloud/android-developers

Forgot to create this PR too, bound to https://github.com/owncloud/android/pull/1493 . Feel free to review & comment.

### DIFF
--- a/src/com/owncloud/android/lib/common/network/FileRequestEntity.java
+++ b/src/com/owncloud/android/lib/common/network/FileRequestEntity.java
@@ -26,6 +26,7 @@
 package com.owncloud.android.lib.common.network;
 
 import java.io.File;
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.io.RandomAccessFile;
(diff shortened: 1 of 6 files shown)


AndyScherzinger:
Looks good to me :+1:

davivel:
Since I was the last one pushing to this branch, @malkomich should review the code.

src/com/owncloud/android/lib/common/network/FileRequestEntity.java
```diff
@@ -124,14 +127,39 @@ public void writeRequest(final OutputStream out) throws IOException {
...
+        } catch (WriteException we) {
+            throw we.getWrapped();
+
         } finally {
-            channel.close();
-            raf.close();
+            try {
+                channel.close();
+                raf.close();
+            } catch (IOException io) {
```
malkomich: Maybe could be cleanest check if Files are null to close them instead of a try-catch treatment.

davivel: No, a null check does not cover the same error, this check is not about that. The method ```close()``` may throw ```IOException``` while working, so we need to handle it or throw it up. In this case we explicitly ignore it because there is not much we can do about it, and because at this point the file was already uploaded or failed-and-was-handled previously in other point.

malkomich:
:+1:

PR #114, opened 2016-03-15
Allow special chars in password

LukeOwncloud:
Fixes https://github.com/owncloud/android/issues/1509
Fixes https://github.com/owncloud/android/issues/1322
Fixes https://github.com/owncloud/core/issues/7894


malkomich:
@LukeOwncloud we can close this issue and track it in #119.
Your PR is right and is fixing the problem, but the old config of Travis is making the tests fail.

PR #116, opened 2016-04-07
[WIP] Update upload operations to use inputstream instances

malkomich:
Related issues:

- Upload via "share with owncloud" from third party app does not work #1514
- Error when some apps try to share content #1529

### DIFF
--- a/src/com/owncloud/android/lib/resources/files/UploadRemoteFileOperation.java
+++ b/src/com/owncloud/android/lib/resources/files/UploadRemoteFileOperation.java
@@ -26,8 +26,11 @@ package com.owncloud.android.lib.resources.files;
 
 import java.io.ByteArrayInputStream;
 import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
+import java.nio.channels.FileChannel;
 import java.util.HashSet;
 import java.util.Set;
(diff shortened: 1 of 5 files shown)


davivel:
Repeated fails of Travis due to: https://code.google.com/p/android/issues/detail?id=206022

Seems a problem with the Google repository providing system images to create the emulator; apparently fixed now.

src/com/owncloud/android/lib/resources/files/UploadRemoteFileOperation.java
```diff
@@ -63,20 +66,24 @@
...
 	
 	protected final AtomicBoolean mCancellationRequested = new AtomicBoolean(false);
 	protected Set<OnDatatransferProgressListener> mDataTransferListeners = new HashSet<OnDatatransferProgressListener>();
 
 	protected RequestEntity mEntity = null;
 
 	public UploadRemoteFileOperation(String localPath, String remotePath, String mimeType) {
-		mLocalPath = localPath;
+		try {
+			mInputStream = new FileInputStream(new File(localPath));
```
davivel: This is making fail the automatic tests; before the operation didn't fail until was run, now it's failing in construction.

Let's delay the decision on this just a bit, since there is a chance that we need to roll back this change fully; content:// URIs, in a general sense, are not apropriate to be stored.

malkomich:
Rollback needed

PR #117, opened 2016-04-07
Android Studio 2.0 Upgrade (build tools 23.0.3 + gradle 2.10)

AndyScherzinger:
The new Android Studio 2.0 uses gradle 2.10 and also comes with a new build tools version.
This is the corresponding library PR for the android app PR https://github.com/owncloud/android/pull/1596

What has been done:

- Update Gradle to 2.10
- Update build tools to 23.0.3
- Updated projects to use Android plugin 2.0.0


davivel:
And having a look here too.

davivel:
Related: https://github.com/owncloud/android/pull/1596

.travis.yml
```diff
@@ -13,6 +13,8 @@ branches:
 before_install:
 - echo no | android create avd --force -n test -t $ANDROID_TARGET --abi $ANDROID_ABI
   -c 20M
+- echo yes | android update sdk --filter extra-android-support --no-ui --force > /dev/null
```
davivel: Why are these needed?

AndyScherzinger: The build process does not update the android support lib artefacts in the local repo and thus upgrading to a support lib version newer than 20.x fails since it can't be found in a maven repo. So this android call updates the support libs so the dependency can be resolved. It is just cosmetics. This can be removed if the gradle file is reverted back to reference support lib v20.0.0

davivel: Travis recommends doing these updates through the ```components``` section.

I'd give it a try to add ``` - extra-android-m2repository``` there, and maybe also ```- extra-android-support```

davivel: Moreover, checking the project, seems there is no real dependency on the support library in this repo; I'll double check in command line; in that case, we should fully drop that dependency.

AndyScherzinger: Okay, I'll check if we can get rid of it completely :)

AndyScherzinger: Yeah! Removed! :) The support lib isn't used at all! Nice catch!

AndyScherzinger:
Code review comments integrated.

AndyScherzinger:
@davivel the PR is imho done since the travis build is working just fine. not sure if there are any tests/checks that need to be done for oC internal/non-pulic build processes. So it tagged it with Ready to test :)

davivel:
Sorry, @AndyScherzinger, but the (main/last) committer of a PR cannot decide alone if contribution is OK from a technical standpoint.

I will resume my review today later.

davivel:
This is OK. :+1:

Tested that works with Gradle in command line (via gradle wrapper in the project), works in Android Studio 2.0, and keeps working in Android Studio 1.5 too.

Thanks a lot, @AndyScherzinger

AndyScherzinger:
As always you're welcome and I am happy to help! :) Sorry for prematurely tagging it ready for test! I misinterpreted the review comment as you being done and overlooked the missing ":+1: comment".

PR #120, opened 2016-04-27
Update gradle android plugin to 2.1.0

AndyScherzinger:
With the new Android Studio 2.1 an new version of the android gradle plugin was made available v2.1.0


davivel:
:+1:

PR #125, opened 2016-06-09
Support to access display name of stored accounts

davivel:


davivel:
Solved, from my POV. Supporting owncloud/android#1709.

Please, @owncloud/android-developers , review.

AndyScherzinger:
Looks good, does the code change imply that username gets replaces by real name? I guess even though I don't know any use case atm there might still be a change for situations for both.

davivel:
Not exactly, the operation was always reading more information than the user name. Moreover, the field we were reading to get the user name was the display name, but for the server environments where was needed, they were always the same.

I will rename the operation, it's misleading.

davivel:
@AndyScherzinger , let me know if it's better now.

AndyScherzinger:
:smiley: :+1:

PR #126, opened 2016-07-21
- changed the buildVersions and target SDK to 24.

GurpreetSK95:
changes made due to problems at [PR #1744](https://github.com/owncloud/android/pull/1744) in OwnCloud/android repo


.travis.yml
```diff
@@ -1,12 +1,14 @@
...
   - tools
-  - build-tools-23.0.3
-  - android-23
-  - sys-img-armeabi-v7a-android-23
+  - build-tools-24.0.0
+  - android-24
+  - sys-img-armeabi-v7a-android-24
 branches:
   only:
   - master
```
davivel: Line 42 needs an update to ```android-24```

davivel:
Hi, @GurpreetSK95 . Did you see my comment about the line 42? Without that change the build will not work. Could you update the PR?

Thanks in advance :)

GurpreetSK95:
@davivel Sorry for the late reply. I just made the changes and hope it works.

davivel:
No problem at all, @GurpreetSK95 :)

Unfortunately, there is another change to do. See the file ```project.properties```; there is another ```android-23``` there that needs to be updated to ```android-24```

GurpreetSK95:
@davivel checked all android-23 occurrences, build tools, files and sdks. Also, checked the details of CI failure. Can you help and tell me why it failed this time?

davivel:
@GurpreetSK95 , in the file ```test_client/tests/project.properties``` you deleted the entry ```-target=android-23``` , but should be replaced by ```-target=android-24```. The automated tests are failing due to that.

In any case, don't worry if the tests fail after that change too. It's expected the test cases fail for PR coming from external repositories due to how Travis and Github handle hidden variables. But when the only problem is that, I can test by myself and override the result.

It's almost done! :)

davivel:
Now it's OK :+1:, but needs to be rebased on master.

If you don't mind, I'll do it by myself and merge.

davivel:
Moved to #130 to rebase. Will merge there.

PR #130, opened 2016-08-12
Changes for Android N

davivel:
From #126, rebased on master.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,14 @@
 sudo: false
 language: android
+jdk:
+  - oraclejdk8
 android:
   components:
   - platform-tools
   - tools
-  - build-tools-23.0.3
-  - android-23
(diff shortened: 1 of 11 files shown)


AndyScherzinger:
Changes look good to me :+1: 
Bumping the target then also means that for Android N compatibility measurements taken by Android won't happen, so thing might break for Android N that worked before on Android N when targeting 23. I don't know of any right away (e.g. Doze is enforced anyways). So I think it is save but for the client app (not the lib), this needs testing.
Just saying since the v22->v23 bump forced me back then to implement the new permission API!

davivel:
Sure. The point is, we need to go there right now. Both to grant instant uploads in N, and to adapt to Doze.

To discuss in owncloud/android#1744, for instance.

AndyScherzinger:
Yes, that is true! :+1:

PR #131, opened 2016-08-17
Updated graddle wrapper and plugin versions from security recommendation

davivel:
Details in https://docs.gradle.org/2.14/release-notes#local-privilege-escalation-when-using-the-daemon


davivel:
cc @owncloud/android-developers 

No functional change, @jesmrec .

AndyScherzinger:
you need to move from mavenCentral() to jcenter() in the build.gradle files, since maven central doesn't have the latest plugin version (yet) and is afaik the new default for Android development.

davivel:
All-in!

jesmrec:
@davivel only internal version changes and the inclusion of jcenter as default repo for gradle. Don't wait if you can merge!! :)

davivel:
🍿

davivel:
Let's go.

PR #141, opened 2016-12-15
Maintain session with server

davivel:


src/com/owncloud/android/lib/common/DynamicSessionManager.java
```diff
@@ -0,0 +1,66 @@
...
+            return mSingleSessionManager.getClientFor(account, context);
+        } else {
+            return mSimpleFactoryManager.getClientFor(account, context);
+        }
+    }
+
+    @Override
+    public OwnCloudClient removeClientFor(OwnCloudAccount account) {
+        OwnCloudClient clientRemoved = mSimpleFactoryManager.removeClientFor(account);
+        OwnCloudClient clientRemoved2 = mSingleSessionManager.removeClientFor(account);
```
davigonz: Using a number in a variable is not such a good idea.If in the future you have more clientRemoved variables, you will continue increasing the number and the variable name will not help to distinguish it from others. You could use instead clientRemovedFactory and clientRemovedSession or something like that.

src/com/owncloud/android/lib/common/DynamicSessionManager.java
```diff
@@ -0,0 +1,66 @@
...
+        if (clientRemoved2 != null) {
```
davigonz: Is there any reason why you are checking clientRemoved2 before than clientRemoved? I would do first the clientRemoved null checking, because this variable has been created before but It's not really important to change... :wink:

davivel: Indeed, there is. In the case that something goes terribly wrong and both variables are not NULL, I expect less  troubles if we return the session-based client and let the caller do whatever it needs to do with it that if we return the other one and prevent the caller do whatever it needs. The factory-based client is all the time creating instance, that are used once an then become trash for the app; the session-based client is not used so happily.

davigonz:
:+1:

PR #142, opened 2016-12-20
Fix crash with invalid TCP port

davivel:
Fixes https://github.com/owncloud/android/issues/1847#issuecomment-267558274

### DIFF
--- a/src/com/owncloud/android/lib/common/OwnCloudClient.java
+++ b/src/com/owncloud/android/lib/common/OwnCloudClient.java
@@ -115,7 +115,7 @@ public class OwnCloudClient extends HttpClient {
                 proxyPort = Integer.parseInt(proxyPortSt);
             }
         } catch (Exception e) {
-            // nothing to do here
+            Log_OC.w(TAG, "Proxy port could not be read, keeping default value " + proxyPort);
         }
 
         if (proxyHost != null && proxyHost.length() > 0) {
@@ -166,11 +166,11 @@ public class OwnCloudClient extends HttpClient {
 
     /**
      * Requests the received method with the received timeout (milliseconds).
-     * <p>
+     *
      * Executes the method through the inherited HttpClient.executedMethod(method).
-     * <p>
+     *
      * Sets the socket and connection timeouts only for the method received.
-     * <p>
+     *
      * The timeouts are both in milliseconds; 0 means 'infinite';
      * < 0 means 'do not change the default'
      *
@@ -200,7 +200,7 @@ public class OwnCloudClient extends HttpClient {
 
     /**
      * Requests the received method.
-     * <p>
+     *
      * Executes the method through the inherited HttpClient.executedMethod(method).
      *
      * @param method HTTP method request.
@@ -361,7 +361,7 @@ public class OwnCloudClient extends HttpClient {
 
     /**
      * Sets the root URI to the ownCloud server.
-     * <p>
+     *
      * Use with care.
      *
      * @param uri
--- a/src/com/owncloud/android/lib/common/OwnCloudClient.java
+++ b/src/com/owncloud/android/lib/common/OwnCloudClient.java
@@ -39,6 +39,7 @@ import org.apache.commons.httpclient.HttpMethodBase;
 import org.apache.commons.httpclient.HttpStatus;
 import org.apache.commons.httpclient.HttpVersion;
 import org.apache.commons.httpclient.URI;
+import org.apache.commons.httpclient.URIException;
 import org.apache.commons.httpclient.cookie.CookiePolicy;
 import org.apache.commons.httpclient.methods.HeadMethod;
 import org.apache.commons.httpclient.params.HttpMethodParams;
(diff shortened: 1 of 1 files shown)


src/com/owncloud/android/lib/common/OwnCloudClient.java
```diff
@@ -54,135 +55,135 @@
...
-    	}
+        String proxyHost = System.getProperty("http.proxyHost");
+        String proxyPortSt = System.getProperty("http.proxyPort");
+        int proxyPort = 0;
+        try {
+            if (proxyPortSt != null && proxyPortSt.length() > 0) {
+                proxyPort = Integer.parseInt(proxyPortSt);
+            }
+        } catch (Exception e) {
+            // nothing to do here
```
davigonz: It's a good practice to log the exceptions if are not thrown, as in this case. If the proxyPort String variable cannot be parsed to int, it will failed, and you are not handle the exception.

src/com/owncloud/android/lib/common/OwnCloudClient.java
```diff
@@ -54,135 +55,135 @@
...
+     * @deprecated Use ExistenceCheckOperation instead
      */
     @Deprecated
     public boolean existsFile(String path) throws IOException, HttpException {
         HeadMethod head = new HeadMethod(getWebdavUri() + WebdavUtils.encodePath(path));
         try {
             int status = executeMethod(head);
             Log_OC.d(TAG, "HEAD to " + path + " finished with HTTP status " + status +
-            		((status != HttpStatus.SC_OK)?"(FAIL)":""));
+                ((status != HttpStatus.SC_OK) ? "(FAIL)" : ""));
```
davigonz: This log is too long and difficult to read, you could set ((status != HttpStatus.SC_OK) ? "(FAIL)" : "")) into a variable before the logging.

src/com/owncloud/android/lib/common/OwnCloudClient.java
```diff
@@ -54,135 +55,135 @@
...
         } finally {
             head.releaseConnection();    // let the connection available for other methods
         }
     }
-    
+
     /**
      * Requests the received method with the received timeout (milliseconds).
-     * 
+     * <p>
```
davigonz: I think that this "< p >" has been mistaken added. There are several along the file.

src/com/owncloud/android/lib/common/OwnCloudClient.java
```diff
@@ -54,135 +55,135 @@
...
+     * @param connectionTimeout Timeout to set for connection establishment
      */
     public int executeMethod(HttpMethodBase method, int readTimeout, int connectionTimeout) throws IOException {
 
         int oldSoTimeout = getParams().getSoTimeout();
         int oldConnectionTimeout = getHttpConnectionManager().getParams().getConnectionTimeout();
         try {
-            if (readTimeout >= 0) { 
+            if (readTimeout >= 0) {
                 method.getParams().setSoTimeout(readTimeout);   // this should be enough...
```
davigonz: Good comments :+1:

src/com/owncloud/android/lib/common/OwnCloudClient.java
```diff
@@ -199,62 +200,85 @@ public int executeMethod(HttpMethodBase method, int readTimeout, int connectionT
...
+        preventCrashDueToInvalidPort(method);
 
-//	        logCookiesAtRequest(method.getRequestHeaders(), "before");
-//	        logCookiesAtState("before");
-            method.setFollowRedirects(false);
+        Log_OC.d(TAG + " #" + mInstanceNumber, "REQUEST " +
+            method.getName() + " " + method.getPath());
 
-            int status = super.executeMethod(method);
+        //logCookiesAtRequest(method.getRequestHeaders(), "before");
```
davigonz: Why do you leave these commented lines? If they are not going to be used anymore, may is a good idea to delete them.

src/com/owncloud/android/lib/common/OwnCloudClient.java
```diff
@@ -199,62 +200,85 @@ public int executeMethod(HttpMethodBase method, int readTimeout, int connectionT
...
 
-//	        logCookiesAtRequest(method.getRequestHeaders(), "after");
-//	        logCookiesAtState("after");
-//	        logSetCookiesAtResponse(method.getResponseHeaders());
+        if (mFollowRedirects) {
+            status = followRedirection(method).getLastStatus();
+        }
 
-            return status;
+        //logCookiesAtRequest(method.getRequestHeaders(), "after");
```
davigonz: Same here

src/com/owncloud/android/lib/common/OwnCloudClient.java
```diff
@@ -199,62 +200,85 @@ public int executeMethod(HttpMethodBase method, int readTimeout, int connectionT
...
         int redirectionsCount = 0;
         int status = method.getStatusCode();
         RedirectionPath result = new RedirectionPath(status, MAX_REDIRECTIONS_COUNT);
         while (redirectionsCount < MAX_REDIRECTIONS_COUNT &&
-                (   status == HttpStatus.SC_MOVED_PERMANENTLY || 
-                    status == HttpStatus.SC_MOVED_TEMPORARILY ||
-                    status == HttpStatus.SC_TEMPORARY_REDIRECT)
-                ) {
-            
+            (status == HttpStatus.SC_MOVED_PERMANENTLY ||
```
davigonz: This OR condition is too long. You could create a variable (boolean HttpStatusOK) with this condition before the while bucle starts and after that, add the variable to the while bucle.

src/com/owncloud/android/lib/common/OwnCloudClient.java
```diff
@@ -365,90 +389,89 @@ public boolean getFollowRedirects() {
...
 
-			// logCookie(cookie);
-		}
+    public String getCookiesString() {
+        Cookie[] cookies = getState().getCookies();
+        String cookiesString = "";
+        for (Cookie cookie : cookies) {
+            cookiesString = cookiesString + cookie.toString() + ";";
+
+            // logCookie(cookie);
```
davigonz: Why do you have a logCookie method and you are not using it?

davivel: To use it in development context when things go really, really bad. In general, we don't want to log them for privacy reasons.

src/com/owncloud/android/lib/common/OwnCloudClient.java
davivel: Will fix

src/com/owncloud/android/lib/common/OwnCloudClient.java
davivel: Thanks :)

src/com/owncloud/android/lib/common/OwnCloudClient.java
davivel: To uncomment it in development context when things go really, really bad. Trust me, let's keep those. See my last comment, too.

src/com/owncloud/android/lib/common/OwnCloudClient.java
davivel: Same answer :)

src/com/owncloud/android/lib/common/OwnCloudClient.java
davigonz: Sure

davigonz:
:+1:

SamuAlfageme:
👍

PR #148, opened 2017-01-25
Update entry point to get user info

davivel:
Fixes wrong path to UserInfo resource that was taking advantage of extra "intelligence" in server version to redirect paths, dropped with OC server 10.


davigonz:
👍

davivel:
ups... overlooked QA again... two in a month... sorry, @jesmrec

PR #151, opened 2017-02-02
Prevent access to accounts of other apps with the same name as an OC …

davivel:
…account

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -28,6 +28,11 @@
     android:versionCode="1"
     android:versionName="1.0" >
 
+    <!-- USE_CREDENTIALS, MANAGE_ACCOUNTS and AUTHENTICATE_ACCOUNTS are needed for API < 23.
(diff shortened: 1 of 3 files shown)


AndroidManifest.xml
```diff
@@ -28,6 +28,11 @@
     android:versionCode="1"
     android:versionName="1.0" >
 
+    <!-- USE_CREDENTIALS, MANAGE_ACCOUNTS and AUTHENTICATE_ACCOUNTS are needed for API < 23.
+        In API >= 23 the do not exist anymore -->
+    <uses-permission android:name="android.permission.USE_CREDENTIALS" />
```
davigonz: This permission may require to show privacy policy

AndroidManifest.xml
```diff
@@ -28,6 +28,11 @@
...
+    <uses-permission android:name="android.permission.AUTHENTICATE_ACCOUNTS" />
```
davigonz: The same as before

PR #152, opened 2017-02-03
Firewall error messages

davigonz:


src/com/owncloud/android/lib/common/operations/ErrorMessageParser.java
```diff
@@ -0,0 +1,138 @@
+/* ownCloud Android Library is available under MIT license
+ *   Copyright (C) 2016 ownCloud GmbH.
```
jabarros: 2017 ;)

src/com/owncloud/android/lib/common/operations/RemoteOperationResult.java
```diff
@@ -242,23 +245,42 @@ public RemoteOperationResult(boolean success, HttpMethod httpMethod) throws IOEx
...
+
+        if (mHttpCode == HttpStatus.SC_FORBIDDEN) {
+            String bodyResponse = httpMethod.getResponseBodyAsString();
+
+            if (bodyResponse != null && bodyResponse.length() > 0) {
+                InputStream is = new ByteArrayInputStream(bodyResponse.getBytes());
+                ErrorMessageParser xmlParser = new ErrorMessageParser();
+                try {
+                    String errorMessage = xmlParser.parseXMLResponse(is);
+                    if (errorMessage != "" && errorMessage != null) {
```
jabarros: You can use isEmpty() method to check it, but if you change it, be sure to change the order:
if(errorMessage != null && !errorMessage.isEmpty())

src/com/owncloud/android/lib/resources/files/DownloadRemoteFileOperation.java
```diff
@@ -160,9 +161,12 @@ protected int downloadFile(OwnCloudClient client, File targetFile) throws HttpEx
                     // TODO some kind of error control!
                 }
 
-            } else {
+            } else if (status != 403){
```
jabarros: Maybe you can define it as a constant:
private static final int ERROR_FORBIDDEN = 403;

PR #157, opened 2017-04-18
Update version of build tools and gradle dependencies

davivel:
To be in line with recommendations of Android Studio 2.3.1, current stable version of the IDE.


davivel:
@jesmrec , @davigonz, no change to test in the library. You might need to update build tools in your environment - 25.0.0 is requested now as a minimum.

PR #161, opened 2017-05-05
Public share warning on capability

davivel:


davivel:
Wrong target branch. Sorry.

PR #162, opened 2017-05-05
Add support for capability to show optional warning about public shares

davivel:


src/com/owncloud/android/lib/resources/status/OCCapability.java
```diff
@@ -294,4 +305,5 @@ public CapabilityBooleanType getFilesVersioning() {
     public void setFilesVersioning(CapabilityBooleanType filesVersioning) {
         this.mFilesVersioning = filesVersioning;
     }
+
```
davigonz: Please delete this line break

davivel: IMHO, this comment is over-zealous. The value of removing this line break is 0, and is not worth the time needed to do it.

PR #163, opened 2017-05-05
Support in library for multiple public shares per file

davivel:


src/com/owncloud/android/lib/resources/shares/CreateRemoteShareOperation.java
```diff
@@ -34,27 +34,57 @@
...
 /**
  * Creates a new share.  This allows sharing with a user or group or as a link.
  */
 public class CreateRemoteShareOperation extends RemoteOperation {
 
     private static final String TAG = CreateRemoteShareOperation.class.getSimpleName();
 
+    private static final String PARAM_NAME = "name";
+    private static final String PARAM_PASSWORD = "password";
+    private static final String PARAM_EXPIRATION_DATE = "expireDate";
```
davigonz: Not important, but this variable could be renamed as expirationDate

src/com/owncloud/android/lib/resources/shares/CreateRemoteShareOperation.java
```diff
@@ -34,27 +34,57 @@
...
     private String mRemoteFilePath;
     private ShareType mShareType;
     private String mShareWith;
-    private boolean mPublicUpload;
+    private boolean mGetShareDetails;
+
+    /**
+     * Name to set for the public link
+     */
+    private String mName = "";
```
davigonz: Is this initialization really needed?

src/com/owncloud/android/lib/resources/shares/UpdateRemoteShareOperation.java
```diff
@@ -150,38 +164,41 @@ public void setPublicUpload(Boolean publicUpload) {
...
         }
         if (mExpirationDateInMillis < 0) {
             // clear expiration date
-            parametersToUpdate.add(new Pair(PARAM_EXPIRATION_DATE, ""));
+            parametersToUpdate.add(new Pair<>(PARAM_EXPIRATION_DATE, ""));
 
         } else if (mExpirationDateInMillis > 0) {
             // set expiration date
-            DateFormat dateFormat = new SimpleDateFormat(FORMAT_EXPIRATION_DATE);
+            DateFormat dateFormat = new SimpleDateFormat(FORMAT_EXPIRATION_DATE, Locale.GERMAN);
```
davigonz: German 🇩🇪 🍺 ? Should it not be Locale.getDefault()

davivel: ```Locale.getDefault()``` would build something based in the language selection in the device. This date is not to be shown to the user, but to send to the server, so we must be sure of using a set of symbols appropriate. Could have been other country, just chose one I expected not to fail.

src/com/owncloud/android/lib/resources/shares/CreateRemoteShareOperation.java
davivel: Better the other way around, I'll change code reference from EXPIRATION to EXIRE. 'expireDate' is the name determined by the OCS Share API for the header, we need to use exactly that.

davigonz: Yeah, you're right

src/com/owncloud/android/lib/resources/shares/UpdateRemoteShareOperation.java
davigonz: 👍

PR #165, opened 2017-06-01
Private link library

davigonz:


src/com/owncloud/android/lib/common/accounts/AccountUtils.java
```diff
@@ -107,6 +109,29 @@ public static String getBaseUrlForAccount(Context context, Account account)
 
 
     /**
+     * Extracts server url for a file
+     * @param context
+     * @param account
+     * @param fileId remote file id
+     * @return url server for file or null on failure
+     */
+    public static String getUrlForFile (Context context, Account account, String fileId)
```
davivel: I would search for another class to place this method, better if it's not in the OC-library. The method implementation might call ```AccountUtils#getBaseUrlForAccount(...)```  if there is no other better way to get the base URL, but responsibilities of ```AccountUtils```  should not include know specific things about individual files.

davigonz: Sorry, not seen this requested change, I've noticed that this class contains many methods for building urls but all of them are more related with accounts instead of with files themselves.  So as you said, it's better to keep this part in the OC-app, not in the library.

davigonz:
PR not needed, no changes in the library to implement private links

PR #166, opened 2017-06-12
Improve feedback when uploading virus files

davigonz:


src/com/owncloud/android/lib/common/operations/RemoteOperationResult.java
```diff
@@ -119,7 +119,8 @@
         DELAYED_FOR_WIFI,
         LOCAL_FILE_NOT_FOUND,
         SERVICE_UNAVAILABLE,
-        SPECIFIC_SERVICE_UNAVAILABLE
+        SPECIFIC_SERVICE_UNAVAILABLE,
+        SPECIFIC_UNSUPPORTED_MEDIA_TYPE
```
davivel: There is no need for the ```SPECIFIC_``` prefix. There is no other 'non-specific' ```UNSUPPORTED_MEDIA_TYPE``` that needs to be distinguished. Let's keep it simple as ```UNSUPPORTED_MEDIA_TYPE```

src/com/owncloud/android/lib/common/operations/RemoteOperationResult.java
```diff
@@ -266,7 +267,27 @@ public RemoteOperationResult(boolean success, HttpMethod httpMethod) throws IOEx
             }
         }
 
-        if (mHttpCode == HttpStatus.SC_SERVICE_UNAVAILABLE) {
+        if (mHttpCode == HttpStatus.SC_UNSUPPORTED_MEDIA_TYPE) {    // 415
+
+            String bodyResponse = httpMethod.getResponseBodyAsString();
```
davivel: Time to refactor this block. The code in this 'if' is repeated several time. Let's define a private method to reduce duplication.

src/com/owncloud/android/lib/common/operations/RemoteOperationResult.java
davigonz: I've used SPECIFIC_... instead of UNSUPPORTED_MEDIA_TYPE just in case we want to handle a common 415 error in the future. Besides, I think that SPECIFIC_UNSUPPORTED_MEDIA_TYPE is more appropriated because we are setting it when the response body contains a **specific** error message.

davigonz:
Added the requested changes @davivel , time to a second review

src/com/owncloud/android/lib/common/operations/RemoteOperationResult.java
davivel: ```just in case we want to ... in the future```

We need to be careful with things that might want or might now want in the future. Names are not something I would prepare "for the future", because they are easy to change if it's needed in the future, and because they can be misleading if no change is needed in the end.

```we are setting it when the response body contains a specific error message```

I would buy this if the name were ```UNSUPPORTED_MEDIA_TYPE_WITH_SPECIFIC_MESSAGE```, but the meaning of ```SPECIFIC_UNSUPPORTED_MEDIA_TYPE``` in plain English is not exactly the same.

Besides, if I'm not wrong we are setting this result if response is 415, no matter if comes with an specific message or not.

davivel: No, I'm wrong in that last point; this code is only set if there is an specific message. The problem is that no special result code is set when there is no specific message.

Let's leave it for the moment, we'll resume the discussion on the naming of ```SPECIFIC_```s when we aboard the generalized storage of remote messages, that has been taken out of the associated issue for now.

PR #168, opened 2017-06-16
Write only share

davigonz:


src/com/owncloud/android/lib/resources/shares/UpdateRemoteShareOperation.java
```diff
@@ -102,6 +102,7 @@ public UpdateRemoteShareOperation(long remoteId) {
         mPassword = null;               // no update
         mExpirationDateInMillis = 0;    // no update
         mPublicUpload = null;
+        mPermissions = OCShare.DEFAULT_PERMISSION;
```
davivel: Why default to this? Why not 0? Is there a way to not specifiy permissions in the operation, expecting that the current value is not changed?

davigonz: See https://github.com/owncloud/android/pull/1988#discussion_r122896761

davivel: :+1:

PR #172, opened 2017-07-20
Update Gradle Plugin to last stable version - 2.3.3

davivel:


davivel:
No functional change, no QA needed.

PR #173, opened 2017-08-04
OAuth2 support

davivel:


src/com/owncloud/android/lib/common/authentication/OwnCloudBasicCredentials.java
```diff
@@ -50,6 +54,8 @@ public OwnCloudBasicCredentials(String username, String password, boolean preemp
 
     @Override
     public void applyTo(OwnCloudClient client) {
+        AuthPolicy.registerAuthScheme(AuthState.PREEMPTIVE_AUTH_SCHEME, BasicScheme.class);
```
davivel: OMG. It was me! Removing...

davivel: OK, constant ```AuthState.PREEMPTIVE_AUTH_SCHEME``` is just "Basic", nothing to do really with preemptive mode of authentication in library; so no harm there.

davivel:
:+1: 

But I cannot approve via Github review because I created the PR. 

@davigonz , will you take a look to the final code?

src/com/owncloud/android/lib/common/operations/RemoteOperation.java
```diff
@@ -25,78 +25,104 @@
...
+
+    /**
+     * When 'true', the operation tries to silently refresh credentials if fails due to lack of authorization.
+     *
+     * Valid for 'execute methods' receiving an {@link Account} instance as parameter, but not for those
+     * receiving an {@link OwnCloudClient}. This is, valid for:
+     *  -- {@link RemoteOperation#execute(Account, Context)}
+     *  -- {@link RemoteOperation#execute(Account, Context, OnRemoteOperationListener, Handler)}
+     */
+    private boolean mSilentRefreshOfAccountCredentials = false;
```
davivel: > We set it as false when methods with no account are executed

Right.

> and as true with methods that have an account available are executed.

No. This set is not done in the object itself, it is done externally, in our app. Other apps using the library could decide not do it, and in that case the refresh wouldn't be done at all.

src/com/owncloud/android/lib/common/operations/RemoteOperation.java
```diff
@@ -166,199 +191,189 @@ public Thread execute(Account account, Context context, OnRemoteOperationListene
                     ("Trying to execute a remote operation with a NULL Context");
         mAccount = account;
         mContext = context.getApplicationContext();
-        mCallerActivity = callerActivity;
-        mClient = null;     // the client instance will be created from mAccount
-                            // and mContext in the runnerThread to create below
+        mClient = null;     // the client instance will be created from
+        // mAccount and mContext in the runnerThread to create below
```
davivel: Fixed.

src/com/owncloud/android/lib/common/operations/RemoteOperation.java
davigonz: > and as true with methods that have an account available are executed.

What I meant with this is that we are setting the object as **true** by using the setter from different parts in the app that execute methods with no account.

Having the variable for other apps that decide not to use the refresh makes sense. 👍

davivel: :+1: then

davivel:
@davigonz, you forgot to approve the review after your last comment; sorry, I need to dismiss it now.

davivel:
Ready for now, merging for beta release.

cc @jesmrec , @davigonz

PR #188, opened 2018-04-25
Upgrade gradle to 3.1.2

davigonz:
Implements https://github.com/owncloud/android/issues/2109


build.gradle
```diff
@@ -1,16 +1,20 @@
...
-repositories {
-    jcenter()
+allprojects {
+    repositories {
+        google()
+        jcenter()
+    }
 }
 
 dependencies {
```
theScrabi: please change 'compile' to 'implementation'.

davigonz: `implementation` is working everywhere except for these cases, it seems that the dependencies are not found and I see the next errors when using implementation and executing `gradlew clean build`

`implementation 'org.apache.jackrabbit:jackrabbit-webdav:2.12.4'`=> Cannot access HttpClient in sampleclient/MainActivity.java

`implementation 'com.squareup.okhttp3:okhttp:3.10.0'`=> package okhttp3 does not exist

davigonz: Hmmm, we should use api instead of compile.

PR #195, opened 2018-10-10
Use UUID v4 to represent request ids and log it

davigonz:
Implements https://github.com/owncloud/android/issues/2055

It needs https://github.com/owncloud/android/pull/2314


davigonz:
This is ready to code review

PR #180, opened 2017-12-17
remove Eclipse structure

hannesa2:
This PR is related to https://github.com/owncloud/android/pull/2076 to make it work. Nothing more.

in submodule I didn't take care about test and sample project, they are in old Eclipse structure and converting them to 'normal' Android structure causes DEX errors,...
This can be done in a additional PR

PR #202, opened 2018-10-25
add missing permission in library

hannesa2:


hannesa2:
This `security/snyk` is maximum annoying
image

michaelstingl:
> This `security/snyk` is maximum annoying

@davigonz @DeepDiver1975 is this tool really helpful?

hannesa2:
@davigonz

hannesa2:
@michaelstingl single point of failure ! 
You need at least two of approver, minimum !

hannesa2:
Sorry I don't found the code, where lint shows me the warning (as far as I remember)
I should always past the lint warning. 
Maybe I was wrong, maybe it disappeared ...

Any way I close it, and when I run into again I make a new one

hannesa2:
and I found more, I will add it too
image

davigonz:
> and I found more, I will add it too

I see... but I am wondering why the app is not failing without those permissions

hannesa2:
@davigonz because app delivers it. 

_So it makes sense to have it in lib and not in app for **other** consumer of this lib_

hannesa2:
@davigonz that means after a merge of this, app can remove this permissions

davigonz:
> that means after a merge of this, app can remove this permissions

Are you sure? `AccountManager` and `ConnectivityManager` are being used in the app well.

hannesa2:
In the future library requires it **and** delivers it. The manifest merger in app will get it from lib, so there is no need for specify in app anymore

davigonz:
> The manifest merger in app will get it from lib, so there is no need for specify in app anymore

Then why the manifest merger is not making the library get the permissions from the app, where are already included? I mean, if we do not specify those permissions in the app but in the library, I suspect the code in the app will appear with some kind of error, same as it appeared in the library.

hannesa2:
Again, currently it's in app, but it should be in lib.
With this PR, other consumer of this lib, get everything out of the box just by including the lib. (**and don't have to add it**)
If you have fears, you can check in future app PR if it works.
This PR only fix providing it in lib

hannesa2:
obsolete now https://github.com/owncloud/android-library/pull/180 includes this as well

PR #223, opened 2018-12-17
Rename library module from owncloud-android-library to lib

davigonz:
Rename library module from owncloud-android-library to lib

Related: https://github.com/owncloud/android/pull/2076#discussion_r242124435


davigonz:
No QA nor code review needed, just renaming the library module

PR #224, opened 2018-12-18
Codestyle

hannesa2:
With this PR every commiter has the same codestyle out of the box (please see `.idea` additions), and so any future PR has hopefully less/no codestyle changes.
I applied codestyle to every file

Now codestyle is per project
image

@davigonz

### DIFF
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -1,0 +1,412 @@
+<component name="ProjectCodeStyleConfiguration">
+  <code_scheme name="Project" version="173">
+    <option name="LINE_SEPARATOR" value="&#10;" />
+    <option name="RIGHT_MARGIN" value="150" />
+    <AndroidXmlCodeStyleSettings>
+      <option name="USE_CUSTOM_SETTINGS" value="true" />
+    </AndroidXmlCodeStyleSettings>
+    <JavaCodeStyleSettings>
+      <option name="FIELD_NAME_PREFIX" value="m" />
+      <option name="STATIC_FIELD_NAME_PREFIX" value="s" />
+      <option name="CLASS_COUNT_TO_USE_IMPORT_ON_DEMAND" value="99999" />
+      <option name="NAMES_COUNT_TO_USE_IMPORT_ON_DEMAND" value="99999" />
+      <option name="PACKAGES_TO_USE_IMPORT_ON_DEMAND">
+        <value />
+      </option>
+      <option name="IMPORT_LAYOUT_TABLE">
+        <value>
+          <package name="android" withSubpackages="true" static="false" />
+          <emptyLine />
+          <package name="" withSubpackages="true" static="false" />
+          <emptyLine />
+          <package name="javax" withSubpackages="true" static="false" />
+          <package name="java" withSubpackages="true" static="false" />
+          <emptyLine />
+          <package name="" withSubpackages="true" static="true" />
+          <emptyLine />
+        </value>
+      </option>
+    </JavaCodeStyleSettings>
+    <JetCodeStyleSettings>
+      <option name="PACKAGES_TO_USE_STAR_IMPORTS">
+        <value>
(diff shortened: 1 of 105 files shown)


hannesa2:
app needs same after merge https://github.com/owncloud/android/pull/2076

DeepDiver1975:
test-client seems to no longer exist in master - the two gradle files have been removed from snyk

davigonz:
Thanks @hannesa2 for this contribution, this is very useful to complete one of the tasks I defined in https://github.com/owncloud/android/issues/2351. 

Can you include an overview about the code style in the original post so that everyone knows it without analyzing the `.idea` file?

hannesa2:
I took one very similar to common, additionally forced brackets and extended line length (we all have modern laptops) and it fits for my 11" Mac Air too.
There are some Kotlin settings toll 
I don't fight for the content/style and everyone can change it, if he knows it better (don't forget to reformat complete project). **Important is, everyone has the same setting**

Btw Android Studio has a bug (I will search for Google ticket), it will always try to change codestyle setting on initial open of project. Workaround is to 

1. un-set codestyle
2. apply 
3. set codestyle
4. apply

hannesa2:
arghhh rebase again https://github.com/owncloud/android-library/pull/226

hannesa2:
This is the bug https://issuetracker.google.com/issues/78097962 please star it at the top left corner by push the star.

image

.idea/codeStyles/Project.xml
```diff
@@ -0,0 +1,412 @@
...
+      <option name="CLASS_COUNT_TO_USE_IMPORT_ON_DEMAND" value="99999" />
+      <option name="NAMES_COUNT_TO_USE_IMPORT_ON_DEMAND" value="99999" />
+      <option name="PACKAGES_TO_USE_IMPORT_ON_DEMAND">
+        <value />
+      </option>
+      <option name="IMPORT_LAYOUT_TABLE">
+        <value>
+          <package name="android" withSubpackages="true" static="false" />
+          <emptyLine />
+          <package name="" withSubpackages="true" static="false" />
```
davigonz: What are these packages with no name used for? There's several ones

.idea/codeStyles/Project.xml
```diff
@@ -0,0 +1,412 @@
+<component name="ProjectCodeStyleConfiguration">
+  <code_scheme name="Project" version="173">
+    <option name="LINE_SEPARATOR" value="&#10;" />
+    <option name="RIGHT_MARGIN" value="150" />
+    <AndroidXmlCodeStyleSettings>
+      <option name="USE_CUSTOM_SETTINGS" value="true" />
+    </AndroidXmlCodeStyleSettings>
+    <JavaCodeStyleSettings>
+      <option name="FIELD_NAME_PREFIX" value="m" />
```
davigonz: How does Android Studio check this when using reformat code? Does it suggest you to add `m` character if is not there?

davigonz:
The code looks good to me @hannesa2 though I put above a couple of questions

.idea/codeStyles/Project.xml
hannesa2: I didn't changed here something special. All is generated by Android Studio.
To be honest, I never investigate what, why, where... it simple works when you use it

.idea/codeStyles/Project.xml
hannesa2: It's this
image

hannesa2: You get no lint warning, when you don't have it.
In my point of view, it's not that important, because the future is Kotlin and don't suggest it

davigonz:
Code approved, @jesmrec , think that checking whether the library project compiles properly and the new code style appears in Android studio preferences would be enough.

michaelstingl:
> Code approved, @jesmrec , think that checking whether the library project compiles properly and the new code style appears in Android studio preferences would be enough.

Works for me 👍

PR #226, opened 2018-12-19
Rename library module from lib to owncloudComLibrary

davigonz:
Rename library module from lib to owncloudComLibrary

It needs https://github.com/owncloud/android/pull/2405


hannesa2:
I'm satisfied with the name, it's very close to my first suggestion 👍

davigonz:
> I'm satisfied with the name, it's very close to my first suggestion 

@hannesa2 yes, thank you for your suggestion, now we can include other future libraries, e.g. with UI stuff or even non ownCloud libraries without confusing names.

PR #231, opened 2019-02-04
Android X

hannesa2:
Only compileSdk increase was needed.

image

Anything else would be a surprise, because Android X is (most) GUI related


davigonz:
@hannesa2 rebase here and we are done

hannesa2:
rebase done

PR #233, opened 2019-02-09
add log level to output

hannesa2:
this is a pre-task to colorize log output and add ability to filter on loglevel


owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/utils/Log_OC.java
```diff
@@ -13,15 +15,18 @@
 public class Log_OC {
     private static final String SIMPLE_DATE_FORMAT = "yyyy/MM/dd HH:mm:ss";
     private static final String LOG_FOLDER_NAME = "log";
-    private static final long MAX_FILE_SIZE = 1000000; // 1MB
+    private static final long MAX_FILE_SIZE = 2000000; // 2MB
```
hannesa2: Because I had so much non-synced files, I've seen only the last sync period (15 minutes)

PR #186, opened 2018-04-13
run code cleanup

theScrabi:
https://github.com/owncloud/android/issues/2153

PR #235, opened 2019-02-14
code cleanup

hannesa2:
Inspired by conflicted and non-maintained https://github.com/owncloud/android-library/pull/186 I did same with this PR, to be able to close https://github.com/owncloud/android-library/pull/186 when this is done

### DIFF
--- a/owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/users/GetRemoteUserAvatarOperation.java
+++ b/owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/users/GetRemoteUserAvatarOperation.java
@@ -58,6 +58,11 @@ public class GetRemoteUserAvatarOperation extends RemoteOperation<GetRemoteUserA
      */
     private int mDimension;
 
+    @Deprecated
+    public GetRemoteUserAvatarOperation(int dimension, String currentEtag) {
+        this(dimension);
+    }
+
     public GetRemoteUserAvatarOperation(int dimension) {
         mDimension = dimension;
     }
--- a/owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/users/GetRemoteUserAvatarOperation.java
+++ b/owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/users/GetRemoteUserAvatarOperation.java
@@ -58,15 +58,8 @@ public class GetRemoteUserAvatarOperation extends RemoteOperation<GetRemoteUserA
      */
     private int mDimension;
 
-    /**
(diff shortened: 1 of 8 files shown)


hannesa2:
Hey guys it's approved since 50 days, what's about to mere it ?

davigonz:
Hi @hannesa2 , we will take it into account in our next planning.

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/users/GetRemoteUserAvatarOperation.java
```diff
@@ -58,15 +58,8 @@
...
     private int mDimension;
 
-    /**
-     * Etag of current local copy of the avatar; if not null, remote avatar will be downloaded only
-     * if its Etag changed.
-     */
-    private String mCurrentEtag;
-
-    public GetRemoteUserAvatarOperation(int dimension, String currentEtag) {
+    public GetRemoteUserAvatarOperation(int dimension) {
```
abelgardep: You removed currentEtag but i can not see if you deleted that parameter in GetRemoteUserAvatarOperation usages. Could you check this @hannesa2 ?

hannesa2: Yes, it needs a change in  app as well.
There are several ways 

1. keep parameter and this noise
2. make PR in app as well and struggle with right time merge of both PRs and keep them in sync and having a temporary pointer to submodule PR (most afford)
3. merge it like it is, and detect a required change, when someone moves to new master and make app-PR then with same step to move submodule pointer

abelgardep: I would discard the third option since we could have some problems in the future.

I would go with the second one because is the way we do it. You can see [here](https://github.com/owncloud/android/pull/2678) a pr from app that needs a pr from library. Same way, in [library pr ](https://github.com/owncloud/android-library/pull/270) we keep a reference to the app pr. This way we move forward both together.

hannesa2: I go with

4. keep old `@Deprecated` and additional new call

So everyone can make a PR in main repo anytime, later or now

abelgardep: Ok, we can move it forward now 👍

jesmrec:
Built correctly. Tested together with `master` in the app module (`c2d0e860`)

Approved.

PR #240, opened 2019-02-25
[New arch] Insert public share for file

davigonz:
Implement https://github.com/owncloud/android/issues/2450
Needs https://github.com/owncloud/android/pull/2461

[QA ]Checks performed after automatic tests: https://github.com/owncloud/android-library/pull/240#issuecomment-479440004


owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/shares/ShareToRemoteOperationResultParser.kt
```diff
@@ -88,13 +88,13 @@ class ShareToRemoteOperationResultParser(private var shareXmlParser: ShareXMLPar
 
             } else if (shareXmlParser?.isWrongParameter!!) {
```
jcmontesmartos: why don't you use when instead of so many else if???

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/shares/ShareParserResult.kt
```diff
@@ -27,4 +27,4 @@ package com.owncloud.android.lib.resources.shares
 
 import java.util.ArrayList
 
-class ShareParserResult(val shares: ArrayList<RemoteShare>, val parserMessage: String)
```
jcmontesmartos: Why don't you use List instead of ArrayList?? IMHO, it's always better to use the interface ;-)

jesmrec:
Some small tests performed here, so the main feature is covered with automatic tests:

- Files:

- [X] Create share with name
- [X] Create share without name
- [X] Create share with password
- [X] Create share with expiration
- [X] Create share with all options
- [X] Create share with error (item not in server anymore)

- Folders:

- [X] Create share with name
- [X] Create share without name
- [X] Create share Download/view
- [X] Create share Download/view/upload
- [X] Create share Upload only
- [x] Create share with password
- [X] Create share with expiration
- [X] Create share with all options
- [X] Create share with error (item not in server anymore)

Capabilities are not supported yet (check https://github.com/owncloud/android/issues/2495)

PR #242, opened 2019-03-07
Android Studio 3.3.2

hannesa2:
see what's changed https://androidstudio.googleblog.com/2019/03/android-studio-332-available.html


davigonz:
@hannesa2 Thanks! I've been dealing with some issues in break points, it seems they have already fixed it for this version

PR #247, opened 2019-04-17
New arch/update public share for file

davigonz:
Implements https://github.com/owncloud/android/issues/2516
Needs https://github.com/owncloud/android/pull/2522

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -9,7 +9,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.3.2'
+        classpath 'com.android.tools.build:gradle:3.4.0'
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlinVersion"
         classpath "org.jetbrains.kotlin:kotlin-allopen:$kotlinVersion"
     }
--- a/owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/shares/ShareXMLParser.kt
+++ b/owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/shares/ShareXMLParser.kt
@@ -179,7 +179,7 @@ class ShareXMLParser {
             } else if (name.equals(NODE_ID, ignoreCase = true)) {// Parse Create XML Response
                 share = RemoteShare()
                 val value = readNode(parser, NODE_ID)
-                share.remoteId = Integer.parseInt(value).toLong()
+                share.id = Integer.parseInt(value).toLong()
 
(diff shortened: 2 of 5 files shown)


build.gradle
```diff
@@ -9,7 +9,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.3.2'
+        classpath 'com.android.tools.build:gradle:3.4.0'
```
jcmontesmartos: brave!! xD

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/shares/ShareXMLParser.kt
```diff
@@ -235,74 +235,74 @@ class ShareXMLParser {
...
+                remoteShare.name = readNode(parser, NODE_NAME)
 
             } else {
                 skip(parser)
             }
         }
 
-        if (isValidShare(share)) {
-            shares.add(share)
+        if (isValidShare(remoteShare)) {
```
jcmontesmartos: Why don't you move this function to RemoteShare class?

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/shares/UpdateRemoteShareOperation.kt
```diff
@@ -151,21 +154,29 @@ class UpdateRemoteShareOperation
...
+            val parser = ShareToRemoteOperationResultParser(
+                ShareXMLParser()
+            )
+            parser.ownCloudVersion = client.ownCloudVersion
+            parser.serverBaseUri = client.baseUri
+            result = parser.parse(putMethod.responseBodyAsString)
+
+            if (result.isSuccess && retrieveShareDetails) {
+                // retrieve more info - PUT only returns the index of the new share
+                val emptyShare = result.data.shares[0]
```
jcmontesmartos: Use first function better
`val emptyShare = result.data.shares.first()`

BTW, I am assuming you are checking shares is not empty, aren't you?

davigonz: Not really, the server returns the updated share with almost no information so I need to retrieve the share with all the information using another operation called GetRemoteShareOperation

build.gradle
jcmontesmartos: You are brave enough to be always in the latest version of Android Studio but you don't want to test @Parcelize, interesting... 🤔

PR #254, opened 2019-06-26
[New Arch] Show + search + create private shares

davigonz:
Implements #2555
Needs https://github.com/owncloud/android/pull/2584


owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/shares/ShareXMLParser.kt
```diff
@@ -281,6 +281,9 @@ class ShareXMLParser {
             } else if (name.equals(NODE_SHARE_WITH_DISPLAY_NAME, ignoreCase = true)) {
                 remoteShare.sharedWithDisplayName = readNode(parser, NODE_SHARE_WITH_DISPLAY_NAME)
 
+            } else if (name.equals(NODE_SHARE_WITH_ADDITIONAL_INFO)) {
```
jcmontesmartos: Can you use here when instead of if, elseif...?

PR #264, opened 2019-08-28
Test bitrise

davigonz:


davigonz:
This is already working, so closing it

PR #266, opened 2019-09-08
Android Studio 3.5.0

hannesa2:
Simply bring library up to date with 
* most recent stable Android Studio
* target to Android Q
* most recent stable Kotlin

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -9,7 +9,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.4.1'
+        classpath 'com.android.tools.build:gradle:3.5.0'
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlinVersion"
         classpath "org.jetbrains.kotlin:kotlin-allopen:$kotlinVersion"
     }
--- a/build.gradle
+++ b/build.gradle
@@ -1,7 +1,7 @@
 buildscript {
     ext {
         // Libraries
-        kotlinVersion = '1.3.21'
+        kotlinVersion = '1.3.50'
     }
 
     repositories {
--- a/owncloudComLibrary/build.gradle
+++ b/owncloudComLibrary/build.gradle
@@ -20,7 +20,7 @@ android {
 
     defaultConfig {
         minSdkVersion 19
-        targetSdkVersion 29
+        targetSdkVersion 28
 
         versionCode = 10000101
         versionName = "1.0.2-beta.1"
--- a/owncloudComLibrary/build.gradle
+++ b/owncloudComLibrary/build.gradle
@@ -16,11 +16,11 @@ allOpen {
 }
 
 android {
-    compileSdkVersion 28
+    compileSdkVersion 29
 
     defaultConfig {
         minSdkVersion 19
-        targetSdkVersion 28
+        targetSdkVersion 29
 
         versionCode = 10000101
         versionName = "1.0.2-beta.1"
--- a/owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/shares/RemoteShare.kt
+++ b/owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/shares/RemoteShare.kt
@@ -104,19 +104,19 @@ class RemoteShare : Parcelable, Serializable {
 
     fun readFromParcel(source: Parcel) {
         id = source.readLong()
-        shareWith = source.readString()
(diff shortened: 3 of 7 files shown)


build.gradle
```diff
@@ -1,15 +1,15 @@
...
+        kotlinVersion = '1.3.50'
     }
 
     repositories {
         google()
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.4.1'
+        classpath 'com.android.tools.build:gradle:3.5.0'
```
theScrabi: Please always check weather the gralde version you pick is compatible with fdroid.

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/shares/RemoteShare.kt
```diff
@@ -104,19 +104,19 @@ class RemoteShare : Parcelable, Serializable {
...
+        name = source.readString().toString()
+        shareLink = source.readString().toString()
         fileSource = source.readLong()
         itemSource = source.readLong()
+        shareType = ShareType.NO_SHARED
         try {
-            shareType = ShareType.valueOf(source.readString())
+            shareType = source.readString()?.let { ShareType.valueOf(it) }
         } catch (x: IllegalArgumentException) {
-            shareType = ShareType.NO_SHARED
```
theScrabi: Is it guaranteed that this case will not occurs in no circumstances?

build.gradle
davigonz: Where can we see the versions that are compatible with fdroid?

owncloudComLibrary/build.gradle
```diff
@@ -16,11 +16,11 @@ allOpen {
 }
 
 android {
-    compileSdkVersion 28
+    compileSdkVersion 29
```
davigonz: Updating Kotlin and Gradle is fine but targetting Android Q is something we tackle carefully since the something could break. We will target Q in the app and the library from a separate issue https://github.com/owncloud/android/issues/2576 , so could you keep 28 here?

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/shares/ShareToRemoteOperationResultParser.kt
```diff
@@ -81,7 +81,9 @@ class ShareToRemoteOperationResultParser(private var shareXmlParser: ShareXMLPar
                             share
                         }
 
-                        result.setData(ShareParserResult(ArrayList(resultData)))
+                        if (resultData != null) {
```
davigonz: You could also write this like:

```
resultData?.let {
    result.data = ShareParserResult(ArrayList(it.toMutableList()))
}
```

build.gradle
hannesa2: I've no clue how to do it

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/shares/RemoteShare.kt
hannesa2: Now it's the default L116 
This way I can get rid of `else` part of `source.readString()?.` L118

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/shares/ShareToRemoteOperationResultParser.kt
hannesa2: Yes I could, but is it necessary or has it an advantage ?

owncloudComLibrary/build.gradle
hannesa2: I reverted target to 28

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/shares/ShareToRemoteOperationResultParser.kt
davigonz: It's not necessary, is just a suggestion

build.gradle
theScrabi: You can find a list of supported gradle versions [here](https://gitlab.com/fdroid/fdroidserver/blob/master/makebuildserver#L149). Apparently 3.5.0 is supported :)

davigonz:
Approved and merged, thanks @hannesa2 for the contribution and @theScrabi for the review.

PR #270, opened 2019-09-30
Add min number of character to search capability

abelgardep:
Implements https://github.com/owncloud/android/issues/2300
Needed for https://github.com/owncloud/android/pull/2678


jesmrec:
Tested with different values of the server `config.php` entry (1..5). Everything correct.

If the `config.php` lacks of the entry, 2 characters is set as default. Also correct.

Approved.

PR #274, opened 2019-11-01
Better logging

hannesa2:
It's about to change logging from self-made to [Timber](https://github.com/JakeWharton/timber).
All is encapsulated with a logging lib which I use in several places.
**I kept to old Log_OC** , to easy migrate and do not touch all files. Especially when you do a huge refactoring

One major advantage for developers:
Before:

image

**NEW**:  you can **click log log line** and jump direct to code ! You see the linenumber as well

image


hannesa2:
Filelogging is still there

hannesa2:
I fixed a synchronization issue

PR #275, opened 2019-11-02
Better logging

hannesa2:
It's about to change logging from self-made to [Timber](https://github.com/JakeWharton/timber).
All is encapsulated with a logging lib which I use in several places.
**I kept to old Log_OC** , to easy migrate and do not touch all files. Especially when you do a huge refactoring

One major advantage for developers:
Before:

image

**NEW**:  you can **click log log line** and jump direct to code ! You see the linenumber as well

image

### DIFF
--- a/owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/utils/Log_OC.java
+++ b/owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/utils/Log_OC.java
@@ -1,212 +1,52 @@
 package com.owncloud.android.lib.common.utils;
 
-import android.util.Log;
+import timber.log.Timber;
 
-import com.owncloud.android.lib.BuildConfig;
-
-import java.io.BufferedWriter;
 import java.io.File;
-import java.io.FileWriter;
-import java.io.IOException;
-import java.text.SimpleDateFormat;
-import java.util.Calendar;
-import java.util.Locale;
 
 public class Log_OC {
-    private static final String SIMPLE_DATE_FORMAT = "yyyy/MM/dd HH:mm:ss";
-    private static final String LOG_FOLDER_NAME = "log";
-    private static final long MAX_FILE_SIZE = 2000000; // 2MB
 
-    private static String mOwncloudDataFolderLog = "owncloud_log";
-
-    private static File mLogFile;
-    private static File mFolder;
-    private static BufferedWriter mBuf;
-
-    private static String[] mLogFileNames = {
-            "currentLog" + BuildConfig.BUILD_TYPE + ".txt",
-            "olderLog" + BuildConfig.BUILD_TYPE + ".txt"
-    };
-
-    private static boolean isMaxFileSizeReached = false;
-    private static boolean isEnabled = false;
+    private static String mOwncloudDataFolderLog;
 
     public static void setLogDataFolder(String logFolder) {
         mOwncloudDataFolderLog = logFolder;
     }
 
-    public static void i(String TAG, String message) {
-        Log.i(TAG, message);
-        appendLog("I: " + TAG + " : " + message);
+    public static void i(String tag, String message) {
+        Timber.i(message);
     }
 
     public static void d(String TAG, String message) {
-        Log.d(TAG, message);
-        appendLog("D: " + TAG + " : " + message);
+        Timber.d(message);
     }
 
     public static void d(String TAG, String message, Exception e) {
-        Log.d(TAG, message, e);
-        appendLog("D: " + TAG + " : " + message + " Exception : " + e.getStackTrace());
+        Timber.d(e, message);
     }
 
     public static void e(String TAG, String message) {
-        Log.e(TAG, message);
-        appendLog("E: " + TAG + " : " + message);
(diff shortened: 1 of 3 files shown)


owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/utils/Log_OC.java
```diff
@@ -1,212 +1,52 @@
...
     public static void d(String TAG, String message, Exception e) {
-        Log.d(TAG, message, e);
-        appendLog("D: " + TAG + " : " + message + " Exception : " + e.getStackTrace());
+        Timber.d(e, message);
     }
 
     public static void e(String TAG, String message) {
-        Log.e(TAG, message);
-        appendLog("E: " + TAG + " : " + message);
+        Timber.d(message);
```
abelgardep: To keep same format, could we use Timber.e(message) ?

owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/utils/Log_OC.java
```diff
@@ -1,212 +1,52 @@
...
-                    mBuf.close();
-                } catch (IOException e) {
-                    e.printStackTrace();
-                }
-            }
-
-            // Check if current log file size is bigger than the max file size defined
-            if (mLogFile.length() > MAX_FILE_SIZE) {
-                isMaxFileSizeReached = true;
-            }
```
abelgardep: What happens if we lose this check? Could logs be huge or how is logs size controlled now?

owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/utils/Log_OC.java
```diff
@@ -1,212 +1,52 @@
...
-import java.text.SimpleDateFormat;
-import java.util.Calendar;
-import java.util.Locale;
 
 public class Log_OC {
-    private static final String SIMPLE_DATE_FORMAT = "yyyy/MM/dd HH:mm:ss";
-    private static final String LOG_FOLDER_NAME = "log";
-    private static final long MAX_FILE_SIZE = 2000000; // 2MB
-
-    private static String mOwncloudDataFolderLog = "owncloud_log";
```
abelgardep: we are losing this initialization, is it pointless?

owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/utils/Log_OC.java
hannesa2: You are right, good catch

owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/utils/Log_OC.java
hannesa2: ### Before
There were two files. Old and new. The new overwrites the old. So it means overall log were split into two similar size files and you see always the current. 

### New
EveryTime you run into Application create a new file with current date is created/reused. The old keeps there till you manually remove it. 
Btw, the file-logging is in my point of view pointless, you see current logcat (in `release` !) anyways. It was a workaround to see what happens in `release` anyhow

owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/utils/Log_OC.java
hannesa2: It's done in L45
`public static void startLogging(String storagePath)` and [here](https://github.com/owncloud/android/blob/6b989467ced4d47281afff37501cca77d5c0748d/owncloudApp/src/main/java/com/owncloud/android/MainApp.kt#L198-L200)

hannesa2:
> By the way, it could be interesting to deprecate those logs with TAG as parameter, and create new ones without it to use them from now on.

We could, but as more sensefull **second** step, a complete replacement (deletion) of `Log_OC` with `Timber` would make sense

hannesa2:
I keep this `Log_OC` only because of seamless logging change during huge refactoring. 
Without your refactoring, I would immediate remove this facade.

owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/utils/Log_OC.java
abelgardep: Ok, thanks for this explanation! I think this is something to check with @davigonz and @jesmrec

jesmrec: I will check in QA stage. Just one question: does the rotation happen every time you open the app? does this not lead to log file flooding?

hannesa2: No, only if it's really opened. Only a click on the icon does not start app, when it's already open in background.
But when you remove app from task list, it will create a file, but maximum only one per day

hannesa2:
Let's continue there https://github.com/owncloud/android-library/pull/283

PR #279, opened 2019-11-08
[New arch] Error handling

davigonz:
Implements https://github.com/owncloud/android/issues/2699 and https://github.com/owncloud/android/issues/2700
Needed for https://github.com/owncloud/android/pull/2708/


owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/shares/RemoteShare.kt
```diff
@@ -195,4 +195,45 @@ class RemoteShare : Parcelable, Serializable {
...
+        fun fromValue(value: Int): ShareType? {
+            return when (value) {
+                -1 -> NO_SHARED
+                0 -> USER
+                1 -> GROUP
+                3 -> PUBLIC_LINK
+                4 -> EMAIL
+                5 -> CONTACT
+                6 -> FEDERATED
+                else -> null
```
abelgardep: Which is the diference between no shared and null in terms of use? Is it needed to have nullability here?

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/status/RemoteCapability.kt
```diff
@@ -102,3 +104,43 @@ class RemoteCapability {
...
+    val isTrue: Boolean
+        get() = value == 1
+
+    companion object {
+        fun fromValue(value: Int): CapabilityBooleanType? {
+            return when (value) {
+                -1 -> UNKNOWN
+                0 -> FALSE
+                1 -> TRUE
+                else -> null
```
abelgardep: Same here, Which is the diference between unknown and null in terms of use? Is it needed to have nullability here?

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/shares/RemoteShare.kt
davigonz: This comes from time ago, we can do it better, I'm on it

davigonz: Well, I've just pushed a commit cleaning up `RemoteShare.kt` and `RemoteCapability.kt` but we can not get rid of the null cases because `ShareType` is an enum in which `NO_SHARED` has -1 value so if we call the method `fun fromValue(value: Int): ShareType?` with value -1, it should retrieve `NO_SHARED` type.

Besides, we use `NO_SHARED` to initialise the share type of a Remote Share

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/status/RemoteCapability.kt
davigonz: Same but changing `NO_SHARED` for `UNKNOWN`

owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/http/interceptors/HttpInterceptor.java
```diff
@@ -46,26 +46,32 @@
...
+            request = currentRequestInterceptor.intercept(request);
         }
 
         Response response = chain.proceed(request);
 
-        for (ResponseInterceptor interceptor : mResponseInterceptors) {
-            response = interceptor.intercept(response);
+        ListIterator<ResponseInterceptor> responseInterceptorIterator = mResponseInterceptors.listIterator();
+
+        while (responseInterceptorIterator.hasNext()) {
```
jcmontesmartos: I think you would change this code. Let me guess, In a coming PR??

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/shares/RemoteShare.kt
```diff
@@ -37,124 +33,27 @@ import java.io.Serializable
...
-    var fileSource: Long = 0
-    var itemSource: Long = 0
-    var shareType: ShareType? = null
-    var permissions: Int = DEFAULT_PERMISSION
-    var sharedDate: Long = INIT_SHARED_DATE
-    var expirationDate: Long = INIT_EXPIRATION_DATE_IN_MILLIS
-    var isFolder: Boolean = path.endsWith(FileUtils.PATH_SEPARATOR)
-    var userId: Long = 0
-
+data class RemoteShare(
```
jcmontesmartos: This is class is neither parcelable nor serializable, are you sure?

owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/http/interceptors/HttpInterceptor.java
davigonz: Yes

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/shares/RemoteShare.kt
davigonz: Yes, is no longer used, OCShare in the domain will keep being parcelable

PR #286, opened 2019-12-11
End of support for <10 servers

davigonz:
Implements https://github.com/owncloud/android/issues/2636
Needed for https://github.com/owncloud/android/pull/2756


davigonz:
👍 LGTM

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/status/OwnCloudVersion.kt
```diff
@@ -163,10 +119,6 @@ class OwnCloudVersion(version: String) : Comparable<OwnCloudVersion>, Parcelable
         return versionValue
     }
 
-    fun supportsRemoteThumbnails(): Boolean {
```
hannesa2, line 166: @davigonz 
You removed this. Does is mean server v10 doesn't support it anymore. Or did you removed it, because is was nowhere used ?

abelgardep: Since only servers oC10 are supported, there was no sense to keep that check anymore
`const val MINIMUM_SERVER_VERSION_FOR_REMOTE_THUMBNAILS = 0x07080000 // 7.8.0`

hannesa2: But why I get a 404 ? https://github.com/owncloud/android/pull/1780#issuecomment-653376761 My server is 10.0.10

PR #295, opened 2020-01-02
fix a typo and some lint warnings

hannesa2:
These are just minor issues

### DIFF
--- a/owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/operations/RemoteOperation.java
+++ b/owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/operations/RemoteOperation.java
@@ -16,6 +16,7 @@ import okhttp3.OkHttpClient;
 
 import java.io.IOException;
 
+@SuppressWarnings("WeakerAccess")
 public abstract class RemoteOperation<T> implements Runnable {
 
     /**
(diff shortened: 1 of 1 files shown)


owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/operations/RemoteOperation.java
```diff
@@ -121,7 +121,7 @@ public Thread execute(OwnCloudClient client,
         if (listener == null) {
             throw new IllegalArgumentException
                     ("Trying to execute a remote operation asynchronously " +
-                            "without a listener to notiy the result");
+                            "without a listener to notify the result");
```
abelgardep: This can be written within the same line, don't you think?

Btw, there are more lines in this file that can be written just in one line.

hannesa2: Of course, and not only in this file, but this is out of scope of this pull request. Keep the focus to the commit message

jesmrec:
No QA needed here

davigonz:
Merged, thanks @hannesa2 , I will update the library reference in the app via https://github.com/owncloud/android/pull/2777

PR #298, opened 2020-01-16
Use Moshi to parse Capabilities

abelgardep:
As this is the common json response, I've created a CommonResponse which will parse any type of response. This response will be used if we finally decide to use moshi as parser. For example, if we use moshi to parse shares, you just need to create a ShareResponse that will be included inside data. This way, ShareXMLParser will be removed, ShareParserResult will be replaced with ShareResponse. This way, RemoteShare could be removed as well as [RemoteCapability](https://github.com/owncloud/android/issues/2716#issuecomment-575281535)?

The same way, [Sharees](https://github.com/owncloud/android/issues/2704) could be parsed creating a ShareeResponse.

```
{
  "ocs": {
    "meta": {
      "status": "ok",
      "statuscode": 100,
      "message": null,
      "totalitems": "",
      "itemsperpage": ""
    },
    "data": [
      {
       ...
      }
    ]
  }
}
```
Implements https://github.com/owncloud/android/issues/2716


owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/status/GetRemoteCapabilitiesOperation.kt
```diff
@@ -45,211 +50,39 @@ import timber.log.Timber
...
-            val requestUri = client.baseUri
-            val uriBuilder = requestUri.buildUpon()
-            uriBuilder.appendEncodedPath(OCS_ROUTE)    // avoid starting "/" in this method
-            uriBuilder.appendQueryParameter(PARAM_FORMAT, VALUE_FORMAT)
-
-            val getMethod = GetMethod(URL(uriBuilder.build().toString()))
-
-            getMethod.addRequestHeader(OCS_API_HEADER, OCS_API_HEADER_VALUE)
-
+            val uriBuilder = client.baseUri.buildUpon().run {
```
davigonz, line 59: Why do you use `apply` below and `run` here? In both cases you want to return the object itself, so I would use `apply` for both. `run` fits better when you want to do something else with the receiver and return a result.

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/status/GetRemoteCapabilitiesOperation.kt
```diff
@@ -45,211 +50,39 @@ import timber.log.Timber
...
-                        Timber.d("*** Added $NODE_FILES")
-                    }
-                }
-                // Result
                 result = RemoteOperationResult(OK)
-                result.data = capability
+                result.data = commonResponse?.ocs?.data?.toRemoteCapability()
 
-                Timber.d("*** Get Capabilities completed ")
+                Timber.d("*** Get Capabilities completed and parsed to ${result.data}")
```
davigonz, line 82: Just curious, why these three *** in some logs?

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/users/GetRemoteUserInfoOperation.kt
```diff
@@ -29,7 +29,8 @@ import com.owncloud.android.lib.common.http.methods.nonwebdav.GetMethod
 import com.owncloud.android.lib.common.operations.RemoteOperation
 import com.owncloud.android.lib.common.operations.RemoteOperationResult
 import com.owncloud.android.lib.common.operations.RemoteOperationResult.ResultCode
-import com.owncloud.android.lib.resources.CommonOcsResponse
+import com.owncloud.android.lib.resources.response.CommonOcsResponse
```
davigonz, line 32: Do you add these two imports but not use them in this commit? How did you get UserInfoResponse before adding its import?

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/response/CapabilityResponse.kt
```diff
@@ -0,0 +1,161 @@
...
+)
+
+@JsonClass(generateAdapter = true)
+data class FileSharingFederation(
+    val incoming: Boolean,
+    val outgoing: Boolean
+)
+
+@JsonClass(generateAdapter = true)
+data class FilesCapabilities(
```
davigonz, line 148: FileCapabilities better?

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/status/GetRemoteCapabilitiesOperation.kt
abelgardep: Not sure, i just left it the way it was. These three *** are included in other remote operations logs.

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/status/GetRemoteCapabilitiesOperation.kt
abelgardep: My bad, updated 👍

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/users/GetRemoteUserInfoOperation.kt
abelgardep: I just moved CommonOcsResponse to another folder. I don't understand your question

PR #300, opened 2020-01-28
Implement OAuth2 with AppAuth

davigonz:
Implements https://github.com/owncloud/android/issues/2793
It needs https://github.com/owncloud/android/pull/2800


davigonz:
@abelgardep ready to code review

owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/authentication/oauth/OAuthConnectionBuilder.java
```diff
@@ -0,0 +1,38 @@
...
+     */
+    public static final OAuthConnectionBuilder INSTANCE = new OAuthConnectionBuilder();
+
+    private static final int CONNECTION_TIMEOUT_MS = (int) TimeUnit.SECONDS.toMillis(15);
+    private static final int READ_TIMEOUT_MS = (int) TimeUnit.SECONDS.toMillis(10);
+
+    @NonNull
+    @Override
+    public HttpURLConnection openConnection(@NonNull Uri uri) throws IOException {
+        Preconditions.checkNotNull(uri, "url must not be null");
```
abelgardep: This NonNull annotations wouldn't be needed in a kotlin file, don't you think❓

owncloudComLibrary/build.gradle
```diff
@@ -24,6 +25,8 @@ android {
 
         versionCode = 10000401
         versionName = "1.0.4.1"
+
+        manifestPlaceholders = [appAuthRedirectScheme: 'oc']
```
abelgardep: Is this also needed here?

gradle.properties
```diff
@@ -0,0 +1,3 @@
+android.enableJetifier=true
+android.useAndroidX=true
+org.gradle.jvmargs=-Xmx1536M
```
abelgardep: 😮

owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/authentication/oauth/OAuthConnectionBuilder.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+public final class OAuthConnectionBuilder implements ConnectionBuilder {
+
+    /**
+     * The singleton instance of the default connection builder.
+     */
+    public static final OAuthConnectionBuilder INSTANCE = new OAuthConnectionBuilder();
+
+    private static final int CONNECTION_TIMEOUT_MS = (int) TimeUnit.SECONDS.toMillis(15);
+    private static final int READ_TIMEOUT_MS = (int) TimeUnit.SECONDS.toMillis(10);
```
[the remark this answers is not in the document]
davigonz: I was thinking of doing this on coming new arch (login) PRs but yes, I will create it as Kotlin.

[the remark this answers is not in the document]
davigonz: What do you mean with `TimeUnit.MILLISECONDS?`

owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/authentication/oauth/OAuthConnectionBuilder.java
davigonz: This file will be kotlin

owncloudComLibrary/build.gradle
davigonz: Yes, otherwise the project won't compile

owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/authentication/oauth/OAuthConnectionBuilder.java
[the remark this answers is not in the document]
abelgardep: Yes, i don't see the point of using `TimeUnit.SECONDS.toMillis()` if there is already a `TimeUnit.MILLISECONDS`, is there a good reason to use that?

owncloudComLibrary/build.gradle
davigonz: I tried what that issue suggests but it doesn't work

gradle.properties
davigonz: Without that file you will see the duplicate classes error below when running `./gradlew ":owncloudComLibrary:assembleAndroidTest" ":owncloudComLibrary:assembleDebug" ":owncloudComLibrary:assembleDebugAndroidTest" ":owncloudComLibrary:assembleDebugUnitTest" ":owncloudComLibrary:assembleRelease" ":owncloudComLibrary:assembleReleaseUnitTest"` command (this is executed on Bitrise)

Screenshot 2020-02-06 at 14 14 50

owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/authentication/oauth/OAuthConnectionBuilder.java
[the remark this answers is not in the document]
davigonz: We can not use `TimeUnit.MILLISECONDS` the way you suggest and writing `private static final int CONNECTION_TIMEOUT_MS = 15000` is even worse imho.

owncloudComLibrary/build.gradle
davigonz: Guess we can not do much more here...

owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/authentication/oauth/OAuthConnectionBuilder.kt
```diff
@@ -0,0 +1,31 @@
...
+import java.util.concurrent.TimeUnit
+
+/**
+ * Based on [net.openid.appauth.connectivity.DefaultConnectionBuilder] but permitting http connections in addition
+ * to https connections
+ */
+class OAuthConnectionBuilder : ConnectionBuilder {
+    @Throws(IOException::class)
+    override fun openConnection(uri: Uri): HttpURLConnection {
+        Preconditions.checkNotNull(uri, "url must not be null")
```
abelgardep: I wouldn't add that check not null since if uri is null, kotlin will throw an exception. I don't know if that method performs aditional checks, so lets move this forward.

PR #301, opened 2020-02-04
Fix multiaccount with OAuth does not work + Owncloud Client factories cleanup

davigonz:
Implements https://github.com/owncloud/android/issues/2805
Needed for https://github.com/owncloud/android/pull/2806


owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/OwnCloudClient.java
```diff
@@ -127,9 +121,6 @@ public int executeHttpMethod(HttpBaseMethod method) throws Exception {
 
     private void checkFirstRedirection(HttpBaseMethod method) {
         final String location = method.getResponseHeader(HttpConstants.LOCATION_HEADER_LOWER);
-        if (location != null && !location.isEmpty()) {
-            mRedirectedLocation = location;
-        }
     }
```
abelgardep: Do we need checkFirstRedirection?

davigonz: AS suggested me to do this because of:

Screenshot 2020-02-05 at 08 40 07

PR #309, opened 2020-02-26
[New arch] Complete login

davigonz:
Implements:
- https://github.com/owncloud/android/issues/2808
- https://github.com/owncloud/android/issues/2794

Needed for https://github.com/owncloud/android/pull/2826

### DIFF
--- a/owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/shares/services/implementation/OCShareService.kt
+++ b/owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/shares/services/implementation/OCShareService.kt
@@ -1,0 +1,91 @@
+/**
+ * ownCloud Android client application
+ *
+ * @author David González Verdugo
+ *
+ * Copyright (C) 2020 ownCloud GmbH.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 2,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ * <p>
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package com.owncloud.android.lib.resources.shares.services.implementation
+
+import com.owncloud.android.lib.common.OwnCloudClient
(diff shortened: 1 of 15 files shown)


owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/shares/services/implementation/OCShareService.kt
```diff
@@ -0,0 +1,91 @@
...
+            remoteFilePath,
+            shareType,
+            shareWith,
+            permissions
+        ).apply {
+            this.name = name
+            this.password = password
+            this.expirationDateInMillis = expirationDate
+            this.publicUpload = publicUpload
+            this.retrieveShareDetails = true
```
abelgardep: `this.` is not required inside an apply, i would remove it.

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/shares/services/implementation/OCShareService.kt
```diff
@@ -0,0 +1,91 @@
...
+    ): RemoteOperationResult<ShareParserResult> =
+        UpdateRemoteShareOperation(
+            remoteId
+        ).apply {
+            this.name = name
+            this.password = password
+            this.expirationDateInMillis = expirationDate
+            this.permissions = permissions
+            this.publicUpload = publicUpload
+            this.retrieveShareDetails = true
```
abelgardep: Same here

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/shares/services/implementation/OCShareeService.kt
```diff
@@ -0,0 +1,42 @@
...
+    ShareeService {
+    override fun getSharees(
+        searchString: String,
+        page: Int,
+        perPage: Int
+    ): RemoteOperationResult<ArrayList<JSONObject>> =
+        GetRemoteShareesOperation(
+            searchString,
+            page,
+            perPage
```
abelgardep: I would add names to call params. It's easier to identify them and it is the way other services do.

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/shares/services/implementation/OCShareService.kt
davigonz: It is needed when we use the same name for the method variables and object properties but I could remove it for `exirationDateInMillis` and `retrieveShareDetails` if you want

abelgardep: Sorry, my bad, you are right. Keep it the way it is, don't worry 👍

PR #311, opened 2020-03-10
Support for usernames with character +

abelgardep:
Implements https://github.com/owncloud/android/issues/2822
Needed for https://github.com/owncloud/android/pull/2836


owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/files/RemoteFile.java
```diff
@@ -115,7 +111,7 @@ public RemoteFile(String path) {
     }
 
     public RemoteFile(final Response davResource, String userId) {
```
davigonz: Creating a RemoteUtil.kt class is not a bad idea but why haven't you converted `RemoteFile.java` to kotlin directly?

abelgardep: Because it may require to perform also some changes to other files. We will migrate this RemoteFile to kotlin in next Synchronization epic for sure 👍

PR #312, opened 2020-03-10
Parse file_source and item_source as string

abelgardep:
Needed for: https://github.com/owncloud/android/pull/2837


owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/shares/ShareXMLParser.kt
```diff
@@ -169,27 +175,29 @@ class ShareXMLParser {
...
-                share!!.shareType = ShareType.PUBLIC_LINK
-                val value = readNode(parser, NODE_URL)
-                share.shareLink = value
-
-            } else if (name.equals(NODE_TOKEN, ignoreCase = true)) {
-                share!!.token = readNode(parser, NODE_TOKEN)
-
-            } else {
-                skip(parser)
+            when {
```
davigonz: 🚀

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/shares/RemoteShare.kt
```diff
@@ -42,8 +42,8 @@ data class RemoteShare(
     var sharedWithAdditionalInfo: String = "",
     var name: String = "",
     var shareLink: String = "",
-    var fileSource: Long = 0,
-    var itemSource: Long = 0,
+    var fileSource: String = "0",
```
davigonz: Why "0" and not ""?

abelgardep: Previously it was 0, so i kept the same value but as String. We are not using them at all, so this is just symbolic..

PR #315, opened 2020-04-17
Open Id Connect along with new arch in login

davigonz:
Implements https://github.com/owncloud/android/issues/2633
Needs https://github.com/owncloud/android/pull/2854


owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/accounts/AccountUtils.java
```diff
@@ -330,6 +330,7 @@ public Account getFailedAccount() {
          * OAuth2 scope
          */
         public static final String KEY_OAUTH2_SCOPE = "oc_oauth2_scope";
+        public static final String OAUTH2_OIDC_SCOPE = "openid offline_access email profile";
```
abelgardep, line 333: Same as i commented in app PR, why are these constants here if they are not used in library?

PR #324, opened 2020-06-04
[WIP] [New Arch] Get Avatar

abelgardep:
Implements https://github.com/owncloud/android/issues/2851


owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/users/GetRemoteUserAvatarOperation.kt
```diff
@@ -0,0 +1,109 @@
...
+import timber.log.Timber
+import java.io.IOException
+import java.io.InputStream
+import java.net.URL
+
+class GetRemoteUserAvatarOperation(private val avatarDimension: Int) :
+    RemoteOperation<RemoteAvatarData>() {
+    override fun run(client: OwnCloudClient): RemoteOperationResult<RemoteAvatarData> {
+        var inputStream: InputStream? = null
+        lateinit var getMethod: GetMethod
```
davigonz, line 42: In my point of view you don't need  to use `lateinit` neither for `getMethod` nor for `result`. `lateinit` makes sense when you create class properties or local variables, but not in a method afak.

And getMethod is immutable => val

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/users/GetRemoteUserAvatarOperation.kt
```diff
@@ -0,0 +1,109 @@
...
+
+        lateinit var result: RemoteOperationResult<RemoteAvatarData>
+        try {
+            val endPoint =
+                client.baseUri.toString() + NON_OFFICIAL_AVATAR_PATH + client.credentials.username + "/" + avatarDimension
```
davigonz, line 47: Constant for "/"?

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/users/GetRemoteUserAvatarOperation.kt
```diff
@@ -0,0 +1,109 @@
...
+            val status = client.executeHttpMethod(getMethod)
+
+            if (isSuccess(status)) {
+                // find out size of file to read
+                val contentLength = getMethod.getResponseHeader(HttpConstants.CONTENT_LENGTH_HEADER).toInt()
+
+                // find out MIME-type!
+                val mimeType = getMethod.getResponseHeader(HttpConstants.CONTENT_TYPE_HEADER)
+
+                if (mimeType == null || !mimeType.startsWith("image")) {
```
davigonz, line 61: I would use a constant for "image" or even a String extension to check mimeTypes => mimeType.isImageType()

abelgardep: That will be done in files rearquitecture.

PR #329, opened 2020-06-24
Use Android separator instead of homemade one

hannesa2:
It's related to https://github.com/owncloud/android/pull/2907

### DIFF
--- a/owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/accounts/AccountUtils.java
+++ b/owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/accounts/AccountUtils.java
@@ -41,6 +41,7 @@ import com.owncloud.android.lib.resources.status.OwnCloudVersion;
 import okhttp3.Cookie;
 import timber.log.Timber;
 
+import java.io.File;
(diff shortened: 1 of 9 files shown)


abelgardep:
Hi  @hannesa2, thanks for this PR. There is still a homemade path separator in [FileUtils](https://github.com/owncloud/android-library/blob/master/owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/files/FileUtils.java) Could you replace it too?

hannesa2:
> Hi @hannesa2, thanks for this PR. There is still a homemade path separator in [FileUtils](https://github.com/owncloud/android-library/blob/master/owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/files/FileUtils.java) Could you replace it too?

@abelgardep It's done

abelgardep:
Bitrise is failing because there are still uses of PATH_SEPARATOR in sample_client, could you check it?

hannesa2:
> Bitrise is failing because there are still uses of PATH_SEPARATOR in sample_client

It's solved now

Btw,  with #304 I would receive an email automatic and there would be no need of an action on your side

PR #335, opened 2020-07-08
[New arch] Fetch list of files

abelgardep:
Implements: https://github.com/owncloud/android/issues/2873
Needed for https://github.com/owncloud/android/pull/2918


owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/files/RemoteFile.java
```diff
@@ -85,6 +85,7 @@ public RemoteFile createFromParcel(Parcel source) {
     private BigDecimal mQuotaUsedBytes;
     private BigDecimal mQuotaAvailableBytes;
     private String mPrivateLink;
+    private String mOwner;
```
davigonz, line 88: I guess this is needed because the account name and user name is not always the same, right?

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/files/ReadRemoteFolderOperation.kt
```diff
@@ -0,0 +1,102 @@
...
+                return RemoteOperationResult<ArrayList<RemoteFile>>(propfindMethod).also {
+                    Timber.w("Synchronized $remotePath ${it.logMessage}")
+                }
+            }
+        } catch (e: Exception) {
+            return RemoteOperationResult<ArrayList<RemoteFile>>(e).also {
+                Timber.e(it.exception, "Synchronized $remotePath")
+            }
+        }
+    }
```
davigonz, line 98:
```suggestion
override fun run(client: OwnCloudClient): RemoteOperationResult<ArrayList<RemoteFile>> =
        try {
            val propfindMethod = PropfindMethod(
                URL(client.userFilesWebDavUri.toString() + WebdavUtils.encodePath(remotePath)),
                DavConstants.DEPTH_1,
                DavUtils.getAllPropset()
            )
            client.setFollowRedirects(true)

            val status = client.executeHttpMethod(propfindMethod)

            if (isSuccess(status)) {
                // Result of the operation
                RemoteOperationResult<ArrayList<RemoteFile>>(ResultCode.OK).apply {
                    val remoteFolder = RemoteFile(propfindMethod.root, AccountUtils.getUserId(mAccount, mContext)).apply {
                        owner = mAccount.name
                    }

                    data.add(remoteFolder)

                    propfindMethod.members.forEach { resource ->
                        val file = RemoteFile(resource, AccountUtils.getUserId(mAccount, mContext))
                        file.owner = mAccount.name
                        data.add(file)
                    }
                }.also {
                    Timber.i("Synchronized $remotePath with ${it.data.size} files. ${it.logMessage}")
                }

            } else { // synchronization failed
                RemoteOperationResult<ArrayList<RemoteFile>>(propfindMethod).also {
                    Timber.w("Synchronized $remotePath ${it.logMessage}")
                }
            }
        } catch (e: Exception) {
            RemoteOperationResult<ArrayList<RemoteFile>>(e).also {
                Timber.e(it.exception, "Synchronized $remotePath")
            }
        }
```

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/files/RemoteFile.java
abelgardep: Not at all, i just want to have it earlier. Previously it was included in FileDataStorageManager, and i think it is easier to handle it here, so localdatasource won't need to ask the current account to add this attribute.

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/files/ReadRemoteFolderOperation.kt
abelgardep: I understand your point of view and it is less verbose, but i think we lose readability using it that way.

PR #350, opened 2020-10-06
New arch/moshi parse shares

theScrabi:
related issue: https://github.com/owncloud/android/issues/2704


abelgardep:
This PR is not valid anymore, it will be merged via #352

Am i right, @theScrabi ?

PR #352, opened 2020-10-14
[New arch] Use moshi to parse sharees

theScrabi:
related issue: https://github.com/owncloud/android/issues/2704
related app pr: https://github.com/owncloud/android/pull/2994


owncloudComLibrary/build.gradle
```diff
@@ -46,4 +54,10 @@ android {
         sourceCompatibility JavaVersion.VERSION_1_8
         targetCompatibility JavaVersion.VERSION_1_8
     }
+
+    testOptions {
+        unitTests {
+            includeAndroidResources = true
+        }
+    }
```
abelgardep, line 62: Do we need this ? 🤔

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/CommonOcsResponse.kt
```diff
@@ -41,6 +42,11 @@ data class OCSResponse<T>(
...
     val status: String,
-    val statuscode: Int,
-    val message: String?
+    @Json(name = "statuscode")
+    val statusCode: Int,
+    val message: String?,
+    @Json(name = "itemsperpage")
+    val itemsPerPage: String?,
+    @Json(name = "totalitems")
+    val totalItems: String?
```
abelgardep, line 51: 🚀 🚀

owncloudComLibrary/src/test/java/com/owncloud/android/lib/ShareeResponseTest.kt
```diff
@@ -0,0 +1,145 @@
...
+        assertEquals(null, response!!.ocs.data.users!![1].value!!.additionalInfo)
+    }
+
+    companion object {
+        val EXAMPLE_RESPONSE = """
+{
+    "ocs": {
+        "data": {
+            "exact": {
+                "groups": [],
```
abelgardep, line 87: I like it, but i think that we could create a new responses package with json files and import those files in the test. That way we will have all responses in the same package as real json and not in companion as strings. What do you think? I could help to do it

owncloudComLibrary/src/test/java/com/owncloud/android/lib/ShareeResponseTest.kt
```diff
@@ -0,0 +1,145 @@
...
+class ShareeResponseTest {
+
+    var response: CommonOcsResponse<ShareeOcsResponse>? = null
+
+    @Before
+    fun prepare() {
+        val moshi = Moshi.Builder().build()
+        val type: Type = Types.newParameterizedType(CommonOcsResponse::class.java, ShareeOcsResponse::class.java)
+        val adapter: JsonAdapter<CommonOcsResponse<ShareeOcsResponse>> = moshi.adapter(type)
+        response = adapter.fromJson(EXAMPLE_RESPONSE)
```
abelgardep, line 24: `response` will be the same response always. Maybe we should also check edge cases, such as empty responses etc. What do you think?

BTW, add license header

owncloudComLibrary/src/test/java/com/owncloud/android/lib/ShareeResponseTest.kt
```diff
@@ -0,0 +1,145 @@
+package com.owncloud.android.lib
```
abelgardep, line 1: From my point of view, tests should be included in the same package as the original but in the test folder, i mean, if `ShareeResponse` is inside package `com.owncloud.android.lib.resources.shares.responses`, this test should be in the same hierarchy and not in the root folder.

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/shares/GetRemoteShareesOperation.kt
```diff
@@ -173,12 +181,6 @@ class GetRemoteShareesOperation
...
         // JSON Node names
-        private const val NODE_OCS = "ocs"
-        private const val NODE_DATA = "data"
-        private const val NODE_EXACT = "exact"
-        private const val NODE_USERS = "users"
-        private const val NODE_GROUPS = "groups"
-        private const val NODE_REMOTES = "remotes"
         const val NODE_VALUE = "value"
         const val PROPERTY_LABEL = "label"
         const val PROPERTY_SHARE_TYPE = "shareType"
```
abelgardep, line 186: Some constants are not used anywhere, could you take a look?

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/shares/GetRemoteShareesOperation.kt
```diff
@@ -78,80 +88,78 @@ class GetRemoteShareesOperation
...
+        val data = ArrayList<JSONObject>() // For result data
+        for (i in 0..jsonResults.size) {
+            for (j in 0 until jsonResults[i].length()) {
+                val jsonResult = jsonResults[i].getJSONObject(j)
+                data.add(jsonResult)
+                Timber.d("*** Added item: ${jsonResult.getString(PROPERTY_LABEL)}")
+            }
+        }
+        return data
+    }
```
abelgardep, line 134: This function is not used anywhere, is it?

abelgardep (changes requested):
Some changes requested @theScrabi

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/shares/responses/ShareeResponse.kt
```diff
@@ -0,0 +1,83 @@
...
+    val label: String,
+    @Json(name = "value")
+    val value: ShareeValue
+)
+
+@JsonClass(generateAdapter = true)
+data class ShareeValue(
+    @Json(name = "shareType")
+    val shareType: Int,
+    @Json(name = "shareWith")
```
abelgardep, line 79: From my point of view, those Json names are redundant. If the json name is equals to variable name, i would not add the `Json(name = "")`

owncloudComLibrary/src/test/java/com/owncloud/android/lib/ShareeResponseTest.kt
```diff
@@ -0,0 +1,145 @@
...
+
+import com.owncloud.android.lib.resources.CommonOcsResponse
+import com.owncloud.android.lib.resources.shares.responses.ShareeOcsResponse
+import com.squareup.moshi.JsonAdapter
+import com.squareup.moshi.Moshi
+import com.squareup.moshi.Types
+import junit.framework.Assert.assertEquals
+import junit.framework.Assert.assertTrue
```
abelgardep, line 9: Use `org.junit.Assert` instead of `junit.framework.Assert` 👍

owncloudComLibrary/src/test/java/com/owncloud/android/lib/ShareeResponseTest.kt
theScrabi: Sounds good

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/shares/GetRemoteShareesOperation.kt
abelgardep: Those properties are not used anymore, you parse it directly using moshi. I think you can remove them.

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/shares/RemoteShare.kt
```diff
@@ -106,17 +106,7 @@ enum class ShareType constructor(val value: Int) {
...
-                1 -> GROUP
-                3 -> PUBLIC_LINK
-                4 -> EMAIL
-                5 -> CONTACT
-                6 -> FEDERATED
-                else -> null
-            }
-        }
+        private val values = values();
+        fun fromValue(value: Int) = values.firstOrNull { it.value == value }
```
abelgardep, line 110:
```suggestion
        fun fromValue(value: Int) = values().firstOrNull { it.value == value }
```

abelgardep (changes requested):
@theScrabi Tiny suggestions

owncloudComLibrary/src/test/java/com/owncloud/android/lib/ShareeResponseTest.kt
abelgardep: Ok, tell me if you need some help to do it

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/shares/GetRemoteShareesOperation.kt
```diff
@@ -173,12 +166,6 @@ class GetRemoteShareesOperation
...
         // JSON Node names
-        private const val NODE_OCS = "ocs"
-        private const val NODE_DATA = "data"
-        private const val NODE_EXACT = "exact"
-        private const val NODE_USERS = "users"
-        private const val NODE_GROUPS = "groups"
-        private const val NODE_REMOTES = "remotes"
         const val NODE_VALUE = "value"
         const val PROPERTY_LABEL = "label"
         const val PROPERTY_SHARE_TYPE = "shareType"
```
abelgardep, line 171: There are some constants that are not used anymore. Please, remove them. @theScrabi

owncloudComLibrary/src/test/resources/com.owncloud.android.lib.resources.sharees.responses/example_sharee_response.json
```diff
@@ -0,0 +1,60 @@
+{
```
abelgardep, line 1: I like it 🎉 🎉  But tests failed, could you take a look? 👍

abelgardep (changes requested):
Take a look at the failing tests and then we move it forward 🚀  @theScrabi

owncloudComLibrary/src/test/java/com/owncloud/android/lib/resources/shares/responses/ShareeResponseTest.kt
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    @Test
+    fun `check structure - ok - contains remotes`() {
+        assertNotEquals(null, response.ocs.data.remotes)
+    }
+
+    @Test
+    fun `check structure - ok - contains users`() {
+        assertNotEquals(null, response.ocs.data.users)
```
abelgardep, line 86: If we are using the same response on each test, why shouldn't we unify them in a single one?

`assertNull`
`assertNotNull` 🤔

owncloudComLibrary/src/test/java/com/owncloud/android/lib/resources/shares/responses/ShareeResponseTest.kt
```diff
@@ -0,0 +1,128 @@
...
+        assertEquals(null, response.ocs.data.users[1].value.additionalInfo)
+    }
+
+    @Test
+    fun `check empty response - ok - parsing ok`() {
+        val moshi = Moshi.Builder().build()
+        val type: Type = Types.newParameterizedType(CommonOcsResponse::class.java, ShareeOcsResponse::class.java)
+        val adapter: JsonAdapter<CommonOcsResponse<ShareeOcsResponse>> = moshi.adapter(type)
+        loadResponses(EMPTY_RESPONSE_JSON, adapter)!!
+    }
```
abelgardep, line 120: To test this, you have a `Before` where you load a EXAMPLE_RESPONSE_JSON and then you load an empty one. What are you testing here? I cannot see any assert 🤔

owncloudComLibrary/src/test/java/com/owncloud/android/lib/resources/shares/responses/ShareeResponseTest.kt
```diff
@@ -0,0 +1,128 @@
...
+    private fun loadResponses(fileName: String, adapter: JsonAdapter<CommonOcsResponse<ShareeOcsResponse>>) =
+        adapter.fromJson(File(fileName).readText())
+
+    @Before
+    fun prepare() {
+        val moshi = Moshi.Builder().build()
+        val type: Type = Types.newParameterizedType(CommonOcsResponse::class.java, ShareeOcsResponse::class.java)
+        val adapter: JsonAdapter<CommonOcsResponse<ShareeOcsResponse>> = moshi.adapter(type)
+        response = loadResponses(EXAMPLE_RESPONSE_JSON, adapter)!!
+    }
```
abelgardep, line 52: I would load responses on each test depending on the testing purpose. You can move adapter to a lateinit var for example, and on each test, simply call: `loadResponses(EXAMPLE_RESPONSE_JSON)` and assert that everything is fine with that response.
```suggestion
    @Before
    fun prepare() {
        val moshi = Moshi.Builder().build()
        val type: Type = Types.newParameterizedType(CommonOcsResponse::class.java, ShareeOcsResponse::class.java)
        val adapter: JsonAdapter<CommonOcsResponse<ShareeOcsResponse>> = moshi.adapter(type)
    }
```

owncloudComLibrary/src/test/java/com/owncloud/android/lib/resources/shares/responses/ShareeResponseTest.kt
```diff
@@ -0,0 +1,128 @@
...
+
+    @Test
+    fun `check structure - ok - example response files exist`() {
+        val file = File(EXAMPLE_RESPONSE_JSON)
+        assertTrue(file.exists())
+    }
```
abelgardep, line 58: From my point of view, this test is not needed. If json do not exists, every test will fail, not only this one.

abelgardep (changes requested):
I have some questions here @theScrabi

owncloudComLibrary/src/test/java/com/owncloud/android/lib/resources/shares/responses/ShareeResponseTest.kt
theScrabi: You are right, but this way you can directly see what happens.

owncloudComLibrary/src/test/java/com/owncloud/android/lib/resources/shares/responses/ShareeResponseTest.kt
theScrabi: Please correct me if I am wrong, but I learned that every test should test one thing and not everything. Am I wrong?

abelgardep (approved):
Approved on my side

PR #354, opened 2020-10-22
Remove cookie persistence

theScrabi:
This will resolve cookie persistence which should solve braking the cookiejar after restarting the app.
App PR: https://github.com/owncloud/android/pull/3002


abelgardep (changes requested):
There are still uses of AccountUtils.Constants.KEY_COOKIES. Please remove it completely 👍

PR #361, opened 2020-12-14
Show logcat live

hannesa2:
It updates logcat content during observation

### QA

- [x] (1) Clear does not work for Logcat https://github.com/owncloud/android-library/pull/361#issuecomment-746211469 [FIXED]

### DIFF
--- a/owncloudComLibrary/build.gradle
+++ b/owncloudComLibrary/build.gradle
@@ -6,7 +6,7 @@ dependencies {
     api 'com.squareup.okhttp3:okhttp:4.6.0'
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlinVersion"
     api 'com.gitlab.ownclouders:dav4android:oc_support_2.1.5'
-    api 'com.github.AppDevNext.Logcat:LogcatCore:2.2.1'
+    api 'com.github.AppDevNext.Logcat:LogcatCore:2.2.2'
 
     // Moshi
     implementation ("com.squareup.moshi:moshi-kotlin:$moshiVersion") {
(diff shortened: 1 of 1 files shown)


abelgardep (approved):
Thanks for updating it 👍

jesmrec:
### (1)

After updating to 2.2.1, the `Clear log` option stop working within the `Logcat` tab. `Logfile` works fine:

Dec-16-2020 13-23-29

@hannesa2 could you take a look?

hannesa2:
> After updating to 2.2.1, the `Clear log` option stop working within the `Logcat` tab. `Logfile` works fine:

The system writes permanent into logcat, and when you clear what do you expect ?
In my point of view a clear log with logcat makes no sense any more. What do you think ?

jesmrec:
I expected same behaviour as `master`:

Dec-16-2020 13-38-56

About stop writing in LogCat, would we lose information about crashes or system regards in problems?  CC @abelgardep

hannesa2:
Thank you for the catch, I fixed it

jesmrec:
Problem fixed. Now, logcat logs are updated "on the fly"

Approved.

Thanks fort he contribution @hannesa2 !!

jesmrec:
After checking again the feature i've realised that after cleaning the Logcat, it stops updating the logs "on live". Is there something wrong @hannesa2 ?

Dec-17-2020 12-35-32

hannesa2:
If there is no new output,  it can't show nothing new.

Do you expect something. In library repo there is s sample app where it logs all the time.
If it works there, it logs on your side as well

hannesa2:
The sample works like this, and yes, it has some permanent background logging

logloop

jesmrec:
ok, what is strange for me is that before cleaning, logs are refreshed every second/two seconds. Then, no log is refreshed as you can see in the GIF. I will take another look, just in case.

jesmrec:
Let's move this forward. Thanks again @hannesa2 !

PR #343, opened 2020-09-22
update cookies and retrive them from middleware

theScrabi:
- updates cookies in the cookie jar instead of overriding them
- runs `GetRemoteStatusOperation` when creating new `OwnCloudClient` in order to retrive cookies from a middleware.
- includes https://github.com/owncloud/android-library/pull/342

Needed for https://github.com/owncloud/android/pull/2970

PR #342, opened 2020-09-15
cleanup GetRemoteStatusOperation

theScrabi:
Cleans GetRemoteStatusOperation which is required for the later bugfixing

____

BUGS & IMPROVEMENTS

- [X] (1) http suffix enforced https://github.com/owncloud/android-library/pull/342#issuecomment-695950683 [FIXED]

PR #368, opened 2021-02-02
[Fix] Cookie handling

abelgardep:
Continuation https://github.com/owncloud/android-library/pull/343

- updates cookies in the cookie jar instead of overriding them
- runs `GetRemoteStatusOperation` when creating new `OwnCloudClient` in order to retrieve cookies from a middleware.
- includes https://github.com/owncloud/android-library/pull/342

Needed for https://github.com/owncloud/android/pull/3074

### DIFF
--- a/owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/SingleSessionManager.java
+++ b/owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/SingleSessionManager.java
@@ -24,8 +24,6 @@
 
 package com.owncloud.android.lib.common;
 
-import android.accounts.Account;
-import android.accounts.AccountManager;
 import android.accounts.AuthenticatorException;
 import android.accounts.OperationCanceledException;
 import android.content.Context;
@@ -37,7 +35,6 @@ import com.owncloud.android.lib.common.http.HttpClient;
 import timber.log.Timber;
 
 import java.io.IOException;
-import java.util.Iterator;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 
@@ -136,7 +133,6 @@ public class SingleSessionManager {
                 Timber.v("reusing client for session %s", sessionName);
             }
 
-            keepCookiesUpdated(context, account, client);
             keepUriUpdated(account, client);
         }
         Timber.d("getClientFor finishing ");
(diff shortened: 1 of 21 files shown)


owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/SingleSessionManager.java
```diff
@@ -111,6 +108,12 @@ public OwnCloudClient getClientFor(OwnCloudAccount account, Context context) thr
                     account.getBaseUri(),
                     context.getApplicationContext(),
                     true);    // TODO remove dependency on OwnCloudClientFactory
+
+            //the next two lines are a hack because okHttpclient is used as a singleton instead of being an
+            //injected instance that can be deleted when required
+            client.clearCookies();
```
theScrabi, line 114: This behavior will change once fix/okhttp_singleton is merged however it's not really colliding with it as that PR will still provide both if these methods. So it does not relay matter here which of the changes is merged first and if these lines are still there then or not. We might however test if they can be removed once fix/okhttp_singleton is merged.

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/status/HttpScheme.kt
```diff
@@ -0,0 +1,32 @@
+/* ownCloud Android Library is available under MIT license
+*   Copyright (C) 2020 ownCloud GmbH.
```
theScrabi, line 2: Maybe we can bump this to 2021

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/status/RemoteServerInfo.kt
```diff
@@ -0,0 +1,30 @@
+/* ownCloud Android Library is available under MIT license
+*   Copyright (C) 2020 ownCloud GmbH.
```
theScrabi, line 2: same here.

owncloudComLibrary/src/test/java/com/owncloud/android/lib/CookieJarImplTest.kt
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    private val cookieStore = HashMap<String, List<Cookie>>().apply {
+        put(SOME_HOST, oldCookies)
+    }
+
+    private val cookieJarImpl = CookieJarImpl(cookieStore)
+
+    @Test
+    fun testContainsCookieWithNameReturnsTrue() {
```
theScrabi, line 34:
```suggestion
    fun `test contains cookie with name - ok`() {
```

Maybe we can use this style of function here to keep consistent with all the other tests.

owncloudComLibrary/src/test/java/com/owncloud/android/lib/CookieJarImplTest.kt
```diff
@@ -0,0 +1,74 @@
...
+        assertTrue(cookieJarImpl.containsCookieWithName(oldCookies, COOKIE_B_OLD.name))
+    }
+
+    @Test
+    fun testContainsCookieWithNameReturnsFalse() {
```
theScrabi, line 39:
```suggestion
    fun `test contains cookie with name - ko`() {
```

theScrabi (changes requested):

owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/http/CookieJarImpl.kt
```diff
@@ -0,0 +1,40 @@
...
+                updatedList.add(oldCookie);
+            }
+        }
+        return updatedList;
+    }
+
+    override fun saveFromResponse(url: HttpUrl, cookies: List<Cookie>) {
+        // Avoid duplicated cookies but update
+        val currentCookies: List<Cookie> = sCookieStore[url.host] ?: ArrayList()
+        val updatedCookies: List<Cookie> = getUpdatedCookies(currentCookies, cookies);
```
abelgardep, line 33: Semi-colons in kotlin :)

owncloudComLibrary/src/test/java/com/owncloud/android/lib/CookieJarImplTest.kt
abelgardep: Great, I will update lists initialization too.

theScrabi (approved):

PR #369, opened 2021-02-02
Android Studio and Gradle update

hannesa2:
Update Android Studio and Gradle

App: https://github.com/owncloud/android/pull/3070


owncloudComLibrary/build.gradle
```diff
@@ -18,11 +18,11 @@ dependencies {
 }
 
 android {
-    compileSdkVersion 29
+    compileSdkVersion 30
```
abelgardep, line 21: We won't update to SDK 30 yet. We need to research and probably apply some changes regarding the scoped storage. We will update the SDK, but not in this PR. Could you drop this commit?

abelgardep (changes requested):
Just one comment @hannesa2

abelgardep (approved):

PR #385, opened 2021-03-24
Fix ktlint findings

hannesa2:


abelgardep (approved):
Thanks for the contribution @hannesa2

PR #391, opened 2021-04-14
ktLint check in library

hannesa2:
It's smarter to check library separately.
When this one is merged, we could use this `.editorconfig` in app as well, to have only one settings file

Don't forget to add `./gradlew ktLintCheck` in CI

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,5 @@
+[*]
+max_line_length = 150
+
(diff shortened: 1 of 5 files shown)


abelgardep (approved):
ktlint running in bitrise. Approved on my side. Thanks for the contribution

jesmrec:
Included in CI, nothing else to check on my side

PR #396, opened 2021-04-30
fix xodo sync bug

theScrabi:
Fix xodo sync bug.
Issue: https://github.com/owncloud/android/issues/2752
App PR: https://github.com/owncloud/android/pull/3220

### DIFF
--- a/owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/files/chunks/ChunkedUploadRemoteFileOperation.java
+++ b/owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/files/chunks/ChunkedUploadRemoteFileOperation.java
@@ -25,7 +25,6 @@
 package com.owncloud.android.lib.resources.files.chunks;
 
 import com.owncloud.android.lib.common.OwnCloudClient;
-import com.owncloud.android.lib.common.http.HttpConstants;
 import com.owncloud.android.lib.common.http.methods.webdav.PutMethod;
 import com.owncloud.android.lib.common.network.ChunkFromFileRequestBody;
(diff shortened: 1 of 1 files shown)


owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/files/chunks/ChunkedUploadRemoteFileOperation.java
```diff
@@ -95,10 +96,6 @@ protected RemoteOperationResult uploadFile(OwnCloudClient client) throws Excepti
             } else {
                 mPutMethod = new PutMethod(new URL(uriPrefix + File.separator + chunkIndex), mFileRequestBody);
 
-                if (mRequiredEtag != null && mRequiredEtag.length() > 0) {
-                    mPutMethod.addRequestHeader(IF_MATCH_HEADER, "\"" + mRequiredEtag + "\"");
-                }
-
```
abelgardep, line 101: Should we remove this header? Isn't it mandatory?

After this removal, there are some unused imports. Clean them 👍

abelgardep (changes requested):
Just a question @theScrabi

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/files/chunks/ChunkedUploadRemoteFileOperation.java
abelgardep: It is not required to add the If-Match header on each chunk, but it should be in the final move according to other clients: https://github.com/owncloud/android/issues/2752#issuecomment-820381240

Issue #378, opened 2021-03-10
Remove redundant use of `HttpClient`


theScrabi:
With the changes done by  [removing the ok http singleton](https://github.com/owncloud/android-library/issues/357) I also introduced a redundend use of the [`HttpClient` wrapper](https://github.com/owncloud/android-library/blob/master/owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/http/HttpClient.java).

The HttpClient gets introduced to the HttpMethod through the constructor:
https://github.com/owncloud/android-library/blob/9d80a95a884715f9d71aa16f86bd36d0d50bce14/owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/http/methods/HttpBaseMethod.kt#L15-L17

However the `execute()` method of the `HttpMethod` is later called from the `OwnCloudClient` which extends the `HttpClient` wrapper: https://github.com/owncloud/android-library/blob/b96a3e9be332544ca0905c4b9f9e179adee514bd/owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/OwnCloudClient.java#L94-L110

Getting "in touch" twice with the client is redundant and should be removed. Here are two possible solutions:
#### 1. Try to Inject the client from the constructor:
    The HttpClient modifying methods in the `HttpBaseMethod` do not need to be changed: https://github.com/owncloud/android-library/blob/9d80a95a884715f9d71aa16f86bd36d0d50bce14/owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/http/methods/HttpBaseMethod.kt#L110
  But the code calling the `method.execute()` in the `OwnCloudClient` has to be moved to the `HttpMethod`: https://github.com/owncloud/android-library/blob/9d80a95a884715f9d71aa16f86bd36d0d50bce14/owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/OwnCloudClient.java#L95

#### 2. Try to inject the client from the execute method:
    The HttpClient modifying methods in the `HttpBaseMethod` need to be modified: https://github.com/owncloud/android-library/blob/9d80a95a884715f9d71aa16f86bd36d0d50bce14/owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/http/methods/HttpBaseMethod.kt#L110
Instead of modifying the OkHttp client directly we could save the modifications and apply them when calling the `execute(client)` method.
The only change that need to be applied to  `OwnCloudClient` would be that `this` is suplied the the `execution()` method. 

------
[...]

PR #427, opened 2021-08-18
remove redundant intercept of httpclient with httpmethod

theScrabi:
Issue: https://github.com/owncloud/android-library/issues/378
App: https://github.com/owncloud/android/pull/3353

____

QA

Reports:

- [ ] (1) Redirections 301 https://github.com/owncloud/android-library/pull/427#issuecomment-930065757

### DIFF
--- a/owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/http/methods/HttpBaseMethod.kt
+++ b/owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/http/methods/HttpBaseMethod.kt
@@ -13,24 +13,39 @@ import java.net.MalformedURLException
 import java.net.URL
 import java.util.concurrent.TimeUnit
 
-abstract class HttpBaseMethod constructor(clientWrapper: HttpClient, url: URL) {
-    var okHttpClient: OkHttpClient
+abstract class HttpBaseMethod constructor(url: URL) {
     var httpUrl: HttpUrl = url.toHttpUrlOrNull() ?: throw MalformedURLException()
     var request: Request
     abstract var response: Response
-
     var call: Call? = null
 
+    var followRedirects: Boolean? = true
+    var retryOnConnectionFailure: Boolean? = false
+    var connectionTimeoutVal: Long? = null
+    var connectionTimeoutUnit: TimeUnit? = null
+    var readTimeoutVal: Long? = null
+    var readTimeoutUnit: TimeUnit? = null
+
     init {
-        okHttpClient = clientWrapper.okHttpClient
         request = Request.Builder()
             .url(httpUrl)
             .build()
     }
 
     @Throws(Exception::class)
-    open fun execute(): Int {
-        return onExecute()
+    open fun execute(httpClient: HttpClient): Int {
+        val okHttpClient = httpClient.okHttpClient.newBuilder().apply {
+            retryOnConnectionFailure?.let { retryOnConnectionFailure(it) }
+            followRedirects?.let { followRedirects(it) }
+            readTimeoutUnit?.let { unit ->
+                readTimeoutVal?.let { readTimeout(it, unit) }
+            }
+            connectionTimeoutUnit?.let { unit ->
+               connectionTimeoutVal?.let { connectTimeout(it, unit) }
+            }
+        }.build()
+
+        return onExecute(okHttpClient)
     }
 
     open fun setUrl(url: HttpUrl) {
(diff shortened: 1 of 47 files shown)


owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/http/methods/HttpBaseMethod.kt
```diff
@@ -13,24 +13,39 @@ import java.net.MalformedURLException
...
-    var okHttpClient: OkHttpClient
+abstract class HttpBaseMethod constructor(url: URL) {
     var httpUrl: HttpUrl = url.toHttpUrlOrNull() ?: throw MalformedURLException()
     var request: Request
     abstract var response: Response
-
     var call: Call? = null
 
+    var followRedirects: Boolean? = true
+    var retryOnConnectionFailure: Boolean? = false
```
abelgardep, line 23: If we use default values, could we remove the nullability?

```suggestion
    var followRedirects: Boolean = true
    var retryOnConnectionFailure: Boolean = false
```

owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/http/methods/HttpBaseMethod.kt
```diff
@@ -13,24 +13,39 @@ import java.net.MalformedURLException
...
+    var connectionTimeoutVal: Long? = null
+    var connectionTimeoutUnit: TimeUnit? = null
+    var readTimeoutVal: Long? = null
+    var readTimeoutUnit: TimeUnit? = null
```
abelgardep, line 27: Just a suggestion that could be out of scope from this PR. 

We could create a new data class that receives the value and the `TimeUnit` and use it here. So we could replace 2 variables with just one, and we won't need nested `?.let` to set the timeouts. Maybe a new issue to do this?

owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/http/methods/webdav/DavMethod.kt
```diff
@@ -27,45 +27,39 @@ import at.bitfire.dav4jvm.Dav4jvm.log
...
 
+    protected abstract fun onDavExecute(davResource: DavOCResource): Int
+
     @Throws(Exception::class)
-    override fun execute(): Int {
+    override fun onExecute(okHttpClient: OkHttpClient): Int {
         return try {
-            onExecute()
+             davResource = DavOCResource(
+                okHttpClient.newBuilder().followRedirects(false).build(),
```
abelgardep, line 57: Why do we set `followRedirects(false)` here? And why follow redirects is true by default if we set it to false every time we perform a davMethod execution? Just curious 🤔

abelgardep (changes requested):
Some questions and suggestions here @theScrabi

owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/http/methods/webdav/DavMethod.kt
theScrabi: dav methods don't allow to have redirects enabled. dav4android will throw if one of its methods get called with a client that hass redirects enabled. therefore this is a last safeguard to disable redirects. Lets for now please keep it like this. ConnectionValidater does change the redirect foo all together so this will not matter much.

owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/http/methods/HttpBaseMethod.kt
```diff
@@ -13,24 +13,39 @@ import java.net.MalformedURLException
...
             .build()
     }
 
     @Throws(Exception::class)
-    open fun execute(): Int {
-        return onExecute()
+    open fun execute(httpClient: HttpClient): Int {
+        val okHttpClient = httpClient.okHttpClient.newBuilder().apply {
+            retryOnConnectionFailure?.let { retryOnConnectionFailure(it) }
+            followRedirects?.let { followRedirects(it) }
```
abelgardep, line 39: Now that they are not nullable, we can get rid of `?.let` here, can't we?
```suggestion
            retryOnConnectionFailure(retryOnConnectionFailure)
            followRedirects(followRedirects)
```

abelgardep (approved):
LGTM

jesmrec:
Let's QA a bit here...

jesmrec:
### (1)

Redirections 301 do not work in comparison with `master` branch. Check out with server `ocis.owncloud.works` that points (301) to `ocis.ocis-traefik.latest.owncloud.works`.

Second request is where difference rises:

master:

`12:02:49 HTTPS GET ocis.owncloud.works /status.php`  -> 301. (Location in response headers) ✅ 
`12:02:55 HTTPS GET ocis.ocis-traefik.latest.owncloud.works /signin/v1/identifier/_/authorize?redirect_uri=oc%3A%2F%2Fandroid.owncloud.com&client_id=e4rAs...`  -> 302  ✅ 

fix/remove_redundant_client_call:

`12:13:30 HTTPS GET  ocis.owncloud.works /status.php` -> 301 (Location in response headers) ✅ 
`12:13:31 HTTPS GET  ocis.ocis-traefik.latest.owncloud.works / ` -> 200 returning an html format in payload ❌ 

showing the error message: `Malformed server configuration`

You can check also with `ocis-a11y.owncloud.works` that points to ocis.a11y.owncloud.works`

theScrabi:
Since the redirect got changed compleatly in favour of the [ConnectionValidator](https://github.com/owncloud/enterprise/issues/4201), it may not make sense to fix this partially. We may need to do the checks @jesmrec did on [that](https://github.com/owncloud/enterprise/issues/4201) branch to.

PR #450, opened 2021-11-21
Connection validator

theScrabi:
This is connection Validator.

App: https://github.com/owncloud/android/pull/3456

## QA

Test plan:

Reports:

- [x] (1) http prefix as default https://github.com/owncloud/android-library/pull/450#issuecomment-993451407
- [x] (2) Location with status https://github.com/owncloud/android-library/pull/450#issuecomment-993458851


owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/http/methods/HttpBaseMethod.kt
```diff
@@ -99,6 +100,11 @@ abstract class HttpBaseMethod constructor(url: URL) {
         return response.body?.byteStream()
     }
 
+    /**
+     * returns the final url after following the last redirect.
+     */
+    open fun getFinalUrl() = response.request.url
```
theScrabi, line 106: one can use this function to fix:
https://github.com/owncloud/android/blob/56a0c79a389ee7a5c1fe7a8d8f78c9f1a153618b/owncloudData/src/main/java/com/owncloud/android/data/authentication/datasources/implementation/OCRemoteAuthenticationDataSource.kt#L53

abelgardep: Not available from an operation scope. Let's check other ways...

abelgardep:
Found some issues connecting to http servers. Further testing required @jesmrec

jesmrec:
### (1)

Behaviour of URL detection is changing when protocol prefix is omitted

Having an `http://` server

Current:

typing `http://192.168.1.23:8080`  ✅ 
typing `192.168.1.23:8080` ❌    -> `Server could not be reached`

Before (expected):

typing `http://192.168.1.23:8080`  ✅ 
typing `192.168.1.23:8080` ✅ 

it is more usual that an `https` connection fallbacks to http in case of non-availability, so, previous behaviour is IMHO the proper one. In case of no prefix, it is assumed the connection is `http`

Pixel2 Android11
`e5e7b933`

jesmrec:
### (2)

About redirections, is finally expected than the `Location` header in response points to `status` in the target redirection URL?

I've checked a couple of servers:

`https://ocis.owncloud.works`. -> 301 pointing to status
`https://ocis-a11y.owncloud.works` -> 301 pointing to URL without status

just to confirm, i guess the correct one is the first one (with status)

theScrabi:
> ### (1)
> 
> Behaviour of URL detection is changing when protocol prefix is omitted
> 
> Having an `http://` server
> 
> Current:
> 
> typing `http://192.168.1.23:8080` white_check_mark typing `192.168.1.23:8080` x -> `Server could not be reached`
> 
> Before (expected):
> 
> typing `http://192.168.1.23:8080` white_check_mark typing `192.168.1.23:8080` white_check_mark
> 
> it is more usual that an `https` connection fallbacks to http in case of non-availability, so, previous behaviour is IMHO the proper one. In case of no prefix, it is assumed the connection is `http`
> 
> Pixel2 Android11 `e5e7b933`

I had to remove the behaviour of outofilling the `http/https` part of the url. Because this required that the app tests if one of the both work. However this would break the connection validator behaviour. At the moment it assumes https if the protocoll is missing from the url. If http was expected connection will fail. This behaviour is similar to the iOS app.

jesmrec:
This issue has taken my attention.

If i build the branch using the Android Studio (debug variant), it builds and works fine.

> Task :owncloudApp:compileDebugSources
> Task :owncloudApp:mergeDebugJavaResource
> Task :owncloudApp:dexBuilderDebug
> Task :owncloudApp:mergeProjectDexDebug
> Task :owncloudApp:packageDebug
> Task :owncloudApp:assembleDebug

BUILD SUCCESSFUL in 1m 21s
151 actionable tasks: 137 executed, 14 up-to-date


If i try to create a signed APK using Android Studio, app does not compile with the following error:

```
/Documents/android/owncloud-android-library/owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/http/HttpClient.java:29: error: package com.facebook.stetho.okhttp3 does not exist
import com.facebook.stetho.okhttp3.StethoInterceptor;
                                  ^
```

any clue about what's going on?

theScrabi:
Ah any stetho refference needs to be removed.

theScrabi:
Ok, so here is the important part:
To disable it to clear cookies on validation this flag needs to be set to `false`:
https://github.com/owncloud/android/blob/a3e09bf2298ecc361c282328d9b8a6ee272d49cd/owncloudApp/src/main/java/com/owncloud/android/dependecyinjection/RemoteDataSourceModule.kt#L68
The default behavior should be false. For the future this value must be brandable, thus this behavior should only be enabled for instances using APM.
In this branch it is enabled since I was developing against an APM.

owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/ConnectionValidator.kt
```diff
@@ -0,0 +1,187 @@
...
+                    singleSessionManager.removeClientFor(account)
+                }
+            }
+            // else: onExecute will finish with status 401
+        }
+        return credentialsWereRefreshed
+    }
+
+    companion object {
+        val VALIDATION_RETRY_COUNT = 3
```
abelgardep, line 185: This could be a `private const val`, right?

owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/ConnectionValidator.kt
```diff
@@ -0,0 +1,187 @@
...
+                            failCounter++
+                        }
+                    } else {
+                        failCounter++
+                        if (contentReply.httpCode == HttpConstants.HTTP_UNAUTHORIZED) {
+                            checkUnauthorizedAccess(client, singleSessionManager, contentReply.httpCode)
+                        }
+                    }
+                }
+                if (successCounter >= failCounter) {
```
abelgardep, line 64: Would it be possible to fail 2 times, and succeed on the third one? What is the expected behaviour in that case?

owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/OwnCloudClient.java
```diff
@@ -92,117 +103,55 @@ public void clearCredentials() {
...
+        }
     }
 
-    private int executeRedirectedHttpMethod(HttpBaseMethod method) throws Exception {
-        boolean repeatWithFreshCredentials;
+    private int saveExecuteHttpMethod(HttpBaseMethod method) throws Exception {
         int repeatCounter = 0;
         int status;
 
+        if(mFollowRedirects) {
```
abelgardep, line 119: Let's reformat the whole file 👍

owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/OwnCloudClient.java
```diff
@@ -92,117 +103,55 @@ public void clearCredentials() {
...
 
-            repeatWithFreshCredentials = checkUnauthorizedAccess(status, repeatCounter);
-            if (repeatWithFreshCredentials) {
-                repeatCounter++;
+            if (!mFollowRedirects &&
+                    !method.getFollowRedirects() &&
+                    mConnectionValidator != null &&
+                    (status == HttpConstants.HTTP_MOVED_TEMPORARILY ||
+                            (!(mCredentials instanceof OwnCloudAnonymousCredentials) &&
+                                    status == HttpConstants.HTTP_UNAUTHORIZED))) {
```
abelgardep, line 145: I would try to simplify or move this to a function. It's kind of hard to follow the conditions 🤔

abelgardep (changes requested):
Some questions and suggestions here @theScrabi

abelgardep (approved):
LGTM

jesmrec:
Some questions about this PR:

the condition to validate is:

```
private boolean shouldConnectionValidatorBeCalled(HttpBaseMethod method, int status) {
        return !mFollowRedirects &&
                !method.getFollowRedirects() &&
                mConnectionValidator != null &&
                (status == HttpConstants.HTTP_MOVED_TEMPORARILY ||
                        (!(mCredentials instanceof OwnCloudAnonymousCredentials) &&
                                status == HttpConstants.HTTP_UNAUTHORIZED));
    } 

```

it returns true if all of them are fulfilled:

- redirection is not followed (yet)
- connection validator instance is not null (already created)
- either request status is 302 (MOVED_TEMPORARILY) or credentials are not anonymous (¿?) and status is not 401 (UNAUTHORIZED)

is this correct?

so, these are the questions:

- "clear cookies" means that the process will just discard the existing ones and get new ones with the same type, right? 
- it will happen only inside the validation process, so the setup parameter `clear_cookies_on_validation` only makes sense if the condition (above) to validate is fulfilled and not in any other cases.
- Well, not easy to test this, unless an apm server is available. How did you check this @theScrabi ?

No regressions in regular servers/authentication methods. Cookies weird behaviour (https://github.com/owncloud/android/issues/2604) will be checked in okhttp Singleton issue once it contains the code written here.

which are the expectations here in terms of customer requirement @michaelstingl? only regression testing as in iOS?

michaelstingl:
> which are the expectations here in terms of customer requirement @michaelstingl? only regression testing as in iOS?

Yes, detailed testing from @s-englert in customer project.

jesmrec:
i will answer myself :)

> "clear cookies" means that the process will just discard the existing ones and get new ones with the same type, right?

yes

> it will happen only inside the validation process, so the setup parameter clear_cookies_on_validation only makes sense if the condition (above) to validate is fulfilled and not in any other cases.

cookie validation happens inside the validation process, so, a regular case is not affected, only apms.

So, we can move this forward, wait for customer feedback and iterate in case of improvements/fix to do.

PR #464, opened 2022-01-31
Fix lint warnings

hannesa2:
It's about to fix these three lint warnings

image


abelgardep (approved):
LGTM, thanks for the contribution

PR #469, opened 2022-02-24
Feature/add stetho

theScrabi:
APP: https://github.com/owncloud/android/pull/3568

This will add facebook stetho and leak canary permanently to the debug version of the app. This makes debuging network requests easier as requests can be viewed all the time without having to use mitm proxy and therefore explicitly change the phone configuration.
These two things will not be pressent in the release build.

### DIFF
--- a/owncloudComLibrary/build.gradle
+++ b/owncloudComLibrary/build.gradle
@@ -16,6 +16,7 @@ dependencies {
 
     testImplementation 'junit:junit:4.13.2'
     testImplementation 'org.robolectric:robolectric:4.7.3'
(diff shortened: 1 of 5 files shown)


abelgardep (approved):

jesmrec:
QA checks:

- [X] Stetho support with some browsers (check note below)
- [X] Canary Leak report
- [X] Release variant not including Stetho
- [X] Release variant not including Canary Leak

NOTE: Stetho is only available for Edge and Brave browsers. Chrome and Chromium does not work fine with the library and Firefox and Opera do not include support. 

With such known restriction, this is OK for me to go out.

Issue #470, opened 2022-03-01
remove sample app from library


theScrabi:
The Sample app in the library is not used anymore, yet we still have to fix it from time to time in order to be able to compile the whole library without compiler issues. We should remove it as we have not user for it anymore.

PR #471, opened 2022-03-01
remove sample client

theScrabi:
This will remove the sample client from the client library.
Issue: https://github.com/owncloud/android-library/issues/470


hannesa2:
Does it makes pain ?
I at least like samples which reduces to a minimum

theScrabi:
In some way it does. We don't test the sample client therefore we can not ensure its proper functionality. You could say development on it is already stall for years. However since it is still there we have to fix compile errors on it from time to time in order to get the whole library through CI.
I think we should remove it, or at best move it to a separate repository. @hannesa2 what do you think?

hannesa2:
I at least build `sample_client` on each pull request properly https://github.com/hannesa2/android-library/runs/5395585442?check_suite_focus=true 

image

For me there is no pain

jesmrec:
The point here is not the "no pain it makes". It is whether the sample app is really useful somehow. It was developed many years ago, it's not maintained and we don't know any user/use case in which it is useful. 

At this point, we need a very good reason not to remove it.

abelgardep (approved):
LGTM

jesmrec:
Nothing to test from my side here... sample app is out. Just don't forget to update the pointer from the app to the library.

PR #496, opened 2022-07-06
Retrieve share type directly within the propfind

abelgardep:
Add a new property to the propfind. So that, we can get if the file is shared directly in just one request.

App: https://github.com/owncloud/android/pull/3711


JuancaG05 (approved):
LGTM

PR #514, opened 2022-10-05
[New arch] Clean legacy classes

abelgardep:
Cleaning up the owncloud version from the owncloud client since it was not used anywhere.

Related to https://github.com/owncloud/android/pull/3765


JuancaG05 (approved):
👍

PR #523, opened 2022-12-15
Unsharing wont return anything anymore since result object was not used

abelgardep:
There is no need to parse the response anymore. The app is not using that data, just checks if it was successful or not.

oC10 returned an empty list
oCIS returns an object

That's why it was not working properly on oCIS


JuancaG05 (approved):
👍

PR #562, opened 2023-04-25
[FEATURE REQUEST] Create new file via Open in Web

JuancaG05:
## Related Issues
Implements https://github.com/owncloud/android/issues/3995

App PR: https://github.com/owncloud/android/pull/4023

### DIFF
--- a/owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/appregistry/CreateRemoteFileWithAppProviderOperation.kt
+++ b/owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/appregistry/CreateRemoteFileWithAppProviderOperation.kt
@@ -1,0 +1,108 @@
+/* ownCloud Android Library is available under MIT license
+ *   Copyright (C) 2023 ownCloud GmbH.
+ *
+ *   Permission is hereby granted, free of charge, to any person obtaining a copy
+ *   of this software and associated documentation files (the "Software"), to deal
+ *   in the Software without restriction, including without limitation the rights
+ *   to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ *   copies of the Software, and to permit persons to whom the Software is
+ *   furnished to do so, subject to the following conditions:
+ *
+ *   The above copyright notice and this permission notice shall be included in
+ *   all copies or substantial portions of the Software.
+ *
+ *   THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+ *   EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ *   MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+ *   NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS
+ *   BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN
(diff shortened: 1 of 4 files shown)


owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/appregistry/CreateRemoteFileWithAppProviderOperation.kt
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    @JsonClass(generateAdapter = true)
+    data class CreateFileWithAppProviderResponse(
+        @Json(name = "file_id")
+        val fileId: String,
+    )
+
+    companion object {
+        private const val TIMEOUT: Long = 5_000
```
abelgardep, line 106: Would be great to unify timeouts across similar operations.

What do you think about that? Do you think it is feasible? 🤔

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/appregistry/CreateRemoteFileWithAppProviderOperation.kt
```diff
@@ -0,0 +1,108 @@
...
+import okhttp3.FormBody
+import okhttp3.RequestBody
+import timber.log.Timber
+import java.net.URL
+import java.util.concurrent.TimeUnit
+
+class CreateRemoteFileWithAppProviderOperation(
+    private val createFileWithAppProviderEndpoint: String,
+    private val parentContainerId: String,
+    private val filename: String,
```
abelgardep, line 47: Should we use camelCase in `fileName`?
```suggestion
    private val fileName: String,
```

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/appregistry/GetUrlToOpenInWebRemoteOperation.kt
```diff
@@ -76,7 +77,7 @@ class GetUrlToOpenInWebRemoteOperation(
         }
     }
 
-    private fun isSuccess(status: Int) = status == HttpConstants.HTTP_OK || status == HttpConstants.HTTP_MULTI_STATUS
+    private fun isSuccess(status: Int) = status == HttpConstants.HTTP_OK
```
abelgardep, line 80: 🥇

abelgardep (commented):
Some suggestions here 🍻

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/appregistry/CreateRemoteFileWithAppProviderOperation.kt
JuancaG05: I used it like that as one single word 😁
https://www.wordreference.com/es/translation.asp?tranword=filename

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/appregistry/CreateRemoteFileWithAppProviderOperation.kt
JuancaG05: Yes, I think that's a good idea, but I see there are lots of timeout variables with different values, and some of them are defined as seconds and others as milliseconds (even sometimes I see 10_000 seconds, I think that's not correct). We would need to review this but I think it's out of the scope of this PR. Feel free to open a new issue with this if you want 🍻

jabarros (approved):
LGTM

